PR #201, opened 2019-10-28
Update/improve sdk loading performance

v4lux:


views/templates/front/paypal.tpl
```diff
@@ -60,6 +47,31 @@
...
+<script>
+/**
+ * Create paypal script
+ */
+function initPaypalScript() {
+  if (typeof paypal !== 'undefined') {
+    return;
+  }
+
+  var psCheckoutScript = document.getElementById('paypalSdkPsCheckout');
```
zalexki: prefer using let

PR #207, opened 2019-10-29
Get shipping cost with tax when HT mode is enabled

v4lux:


classes/GenerateJsonPaypalOrder.php
```diff
@@ -198,7 +199,7 @@ public function createJsonFromData($params)
         // set discount value
         $payload['amount']['breakdown']['discount'] = [
             'currency_code' => $params['currency']['iso_code'],
-            'value' => abs($params['cart']['totals']['total_including_tax']['amount'] - $totalProductsWithoutTax - $totalTax - $params['cart']['subtotals']['shipping']['amount'] - $handlingValue),
+            'value' => abs($params['cart']['totals']['total_including_tax']['amount'] - $totalProductsWithoutTax - $totalTax - $params['cart']['shipping_cost'] - $handlingValue),
```
zalexki: document this 'hack' that has i understand fix the rounding mismatch problem coming from prestashop core vs paypal

PR #213, opened 2019-10-31
Bump version to 1.2.6

v4lux:


Quetzacoalt91:
As the narketplace cannot detect properly the version in the module, shall we remove this constant from the class and write `$this->version = '1.2.6':` instead?

Quetzacoalt91:
Well, the marketplace never took in account this kind of constant properly We're facing an improvement instead of a regression, which makes this issue not critical. And I'm not even sure they are aware of this.

In the meantime we must add some changes in the module, or we will encounter the same issue as during the previous release. :/

ps_checkout.php
```diff
@@ -46,8 +46,11 @@
 
 class Ps_checkout extends PaymentModule
 {
+<<<<<<< Updated upstream
     const VERSION = '1.2.6';
 
+=======
+>>>>>>> Stashed changes
```
Quetzacoalt91: Conflict marks to be removed

v4lux: Thx, solved :)

PR #215, opened 2019-11-14
Update/refactoring

v4lux:


classes/Api/Payment/Order.php
```diff
@@ -65,23 +65,19 @@ public function capture($orderId, $merchantId)
...
         ]);
 
-        if (isset($response['checkoutError']) && $response['checkoutError'] === true) {
+        if (false === $response['status']) {
             return $response;
         }
 
-        if (false === isset($response['purchase_units'][0]['payments']['captures'][0])) {
-            return false;
-        }
```
Quetzacoalt91: Isn't this check still useful? You cannot be sure the information exist.

classes/Builder/Payload/OnboardingPayloadBuilder.php
```diff
@@ -0,0 +1,199 @@
...
+        $this->payload = new Payload();
+    }
+
+    /**
+     * Return the result of the payload
+     * Clean the builder to be ready to producing a new payload
+     *
+     * @return Payload
+     */
+    public function getPayload()
```
Quetzacoalt91: This method is not only a getter as it resets the Patload attribute. I recommend using another verb, like `purgePayload` or `getAndResetPayload`

classes/Builder/Payload/OnboardingPayloadBuilder.php
```diff
@@ -0,0 +1,199 @@
...
+
+    /**
+     * Build base node
+     */
+    public function buildBaseNode()
+    {
+        $language = \Language::getLanguage(\Context::getContext()->employee->id_lang);
+        $locale = $language['locale'];
+
+        $this->payload->items += [
```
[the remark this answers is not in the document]
Quetzacoalt91: Or the classic addition of a value:

```suggestion
        $this->payload->items[] = [
```

classes/Builder/Payload/OrderPayloadBuilder.php
```diff
@@ -0,0 +1,299 @@
...
+            'invoice_id' => '',
+            'description' => $this->truncate('Checking out with your cart from ' . \Configuration::get('PS_SHOP_NAME'), 127),
+            'soft_descriptor' => $this->truncate(\Configuration::get('PS_SHOP_NAME'), 22),
+            'amount' => [
+                'currency_code' => $this->cart['currency']['iso_code'],
+                'value' => $this->cart['cart']['totals']['total_including_tax']['amount'],
+            ],
+            'shipping' => [
+                'name' => [
+                    'full_name' => 'Mr ' . $this->cart['addresses']['shipping']->lastname . ' ' . $this->cart['addresses']['shipping']->firstname,
```
Quetzacoalt91: Hardcoded `Mr`?

classes/Builder/Payload/OrderPayloadBuilder.php
```diff
@@ -0,0 +1,299 @@
...
+                'merchant_id' => (new PaypalAccountRepository())->getMerchantId(),
+            ],
+            'application_context' => [
+                'brand_name' => \Configuration::get('PS_SHOP_NAME'),
+                'shipping_preference' => 'SET_PROVIDED_ADDRESS',
+            ],
+            'roundingConfig' => (string) \Configuration::get('PS_ROUND_TYPE') . '-' . (string) \Configuration::get('PS_PRICE_ROUND_MODE'),
+        ];
+
+        $this->getPayload()->setItems($node);
```
Quetzacoalt91: You may move this call at the end of the method, and avoid duplicate it in each `if`.

classes/Builder/Payload/OrderPayloadBuilder.php
```diff
@@ -0,0 +1,299 @@
...
+        // set handling cost id needed -> principally used in case of gift_wrapping
+        if (isset($this->cart['cart']['subtotals']['gift_wrapping'])) {
+            $handlingValue += $this->cart['cart']['subtotals']['gift_wrapping']['amount'];
+
+            $node['amount']['breakdown']['handling'] = [
+                'currency_code' => $this->cart['currency']['iso_code'],
+                'value' => $handlingValue,
+            ];
+
+            $this->getPayload()->setItems($node);
```
Quetzacoalt91: Unneeded, you already have a setItems at the end of this method

classes/Builder/Payload/Payload.php
```diff
@@ -38,7 +38,14 @@ class Payload
...
 
-    public function setItems($array)
+    /**
+     * Setter for items
+     * Use array_replace_recursive in order to merge the new
+     * content with the previous one.
+     *
+     * @param array $array
+     */
+    public function setItems(array $array)
```
zalexki: it is not a setter anymore then, you should name it something like `addItems` or even maybe `mergeItems`

v4lux: Done ! `addAndMergeItems()` !!

classes/Builder/Payload/OrderPayloadBuilder.php
v4lux: Removed.

classes/Builder/Payload/OrderPayloadBuilder.php
v4lux: Removed :)

classes/Builder/Payload/OrderPayloadBuilder.php
v4lux: Instantiate the gender form the customer and get the gender name.

classes/Builder/Payload/Payload.php
```diff
@@ -0,0 +1,65 @@
...
+class Payload
+{
+    /**
+     * Payload content
+     *
+     * @var array
+     */
+    private $items = [];
+
+    public function setItems($array)
```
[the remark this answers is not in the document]
v4lux: Phpdoc added.

classes/PaypalOrder.php
```diff
@@ -45,13 +45,13 @@ public function __construct($id)
      */
     private function loadOrder($id)
     {
-        $order = (new Order(\Context::getContext()->link))->fetch($id);
+        $response = (new Order(\Context::getContext()->link))->fetch($id);
 
-        if (false === $order) {
-            return false;
+        if (false === $response['status']) {
+            return $response;
```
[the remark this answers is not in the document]
v4lux: Indeed, fixed :)

classes/Builder/Payload/OrderPayloadBuilder.php
```diff
@@ -0,0 +1,313 @@
...
+            $paypalItem['sku'] = $this->truncate($value['unity'], 127);
+            $paypalItem['unit_amount']['currency_code'] = $this->cart['currency']['iso_code'];
+            $paypalItem['unit_amount']['value'] = $value['total'] / $value['quantity'];
+            $paypalItem['tax']['currency_code'] = $this->cart['currency']['iso_code'];
+            $paypalItem['tax']['value'] = ($value['total_wt'] - $value['total']) / $value['quantity'];
+            $paypalItem['quantity'] = $value['quantity'];
+            $paypalItem['category'] = $value['is_virtual'] === '1' ? 'DIGITAL_GOODS' : 'PHYSICAL_GOODS';
+
+            $totalProductsWithoutTax += $paypalItem['unit_amount']['value'] * $value['quantity'];
+            $totalTax += $paypalItem['tax']['value'] * $value['quantity'];
```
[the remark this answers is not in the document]
v4lux: Create two dedicated methods to handle this calculation.

classes/Api/Payment/Onboarding.php
```diff
@@ -31,122 +32,35 @@ class Onboarding extends PaymentClient
...
-     * Get the iso code of the default currency for the shop
-     *
-     * @return string iso code
-     */
-    private function getCurrencyIsoCode()
-    {
-        $currency = \Currency::getCurrency((int) \Configuration::get('PS_CURRENCY_DEFAULT'));
+        if (false === isset($response['body']['links']['1']['href'])) {
+            return $response['status'] = false;
+            return $response;
```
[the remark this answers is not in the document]
zalexki: $response is a homemade array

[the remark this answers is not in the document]
Quetzacoalt91: Yeah this homemade array is the improvement brought by this PR, but that's not the point here. If you look at the controller requesting the onboardling link, then at the VueJs state, I'm not sure they are expecting the whole array as a response.

In the vueJs component, the following check is done: 

```js
getOnboardingLink() {
        if (this.paypalOnboardingLink !== false) {
          this.$refs.paypalButton.click();
          return;
        }
```
In case of error with the PayPal API, the whole response array will be returned from the ajax request, and the status is checked nowhere. That's why I was more expecting a `return false;` here.

classes/Presenter/Store/Modules/PaypalModule.php
```diff
@@ -40,7 +40,7 @@ public function present()
         $paypalModule = [
             'paypal' => [
                 'idMerchant' => $idMerchant,
-                'paypalOnboardingLink' => false,
+                'paypalOnboardingLink' => '',
```
Quetzacoalt91: What if the reason in  change the default value?

v4lux: We need a default value that is not "false". We set this value to false when we cannot retrieve an onboarding link from the api in order to display an error to the merchant.

classes/Builder/Payload/OrderPayloadBuilder.php
```diff
@@ -48,38 +48,51 @@ public function __construct(array $cart)
...
     }
 
     /**
      * Build payload without cart details
+     *
+     * @param bool $expressCheckout Allow to build the payload with more or less content depending if
+     * the customer use expresse checkout or not
      */
-    public function buildMinimalPayload()
+    public function buildMinimalPayload($expressCheckout = false)
```
Quetzacoalt91: You're adding a prameter for a new feature. What will happen when we add, let's say ... 4 more features? We will get our payloads with `$buider->buildMinimalPayload(true, false, false, true)`?

This will be difficult to understand. My advice would be to set these options via setters, in case you absolutely need them everywhere.

For instance:

```php
$builder->toggleExpressCheckout(true);
$builder->buildMinimalPayload();
```

What do you think?

classes/Presenter/Store/Modules/PaypalModule.php
Matt75: When false, an error message is displayed in account-list.vue

classes/Builder/Payload/OrderPayloadBuilder.php
v4lux: Done! :)

classes/Builder/Payload/OrderPayloadBuilder.php
```diff
@@ -48,39 +54,33 @@ public function __construct(array $cart)
...
         $this->buildItemsNode();
 
-        if (!$expressCheckout) {
+        if (false === $this->expressCheckout) {
             $this->buildShippingNode();
             $this->buildPayerNode();
         }
 
-        $this->buildApplicationContextNode($expressCheckout);
+        $this->buildApplicationContextNode($this->expressCheckout);
```
Quetzacoalt91: Parameter to be removed?

v4lux: Indeed, removed and use the attribute.

PR #216, opened 2019-11-18
Fix firebase refresh token

Matt75:
Do not make call without refresh token => if module is not configured or merchant logout from PsAccount
Refresh token date storage now work with multishop

### DIFF
--- a/classes/Api/Firebase/Token.php
+++ b/classes/Api/Firebase/Token.php
@@ -73,7 +73,7 @@ class Token extends FirebaseClient
     {
         $refresh_token = \Configuration::get('PS_PSX_FIREBASE_REFRESH_TOKEN');
 
-        return ! empty($refresh_token);
+        return !empty($refresh_token);
     }
 
     /**
--- a/classes/Api/Firebase/Token.php
+++ b/classes/Api/Firebase/Token.php
@@ -48,9 +48,32 @@ class Token extends FirebaseClient
         if (isset($response['id_token']) && isset($response['refresh_token'])) {
             \Configuration::updateValue('PS_PSX_FIREBASE_ID_TOKEN', $response['id_token']);
             \Configuration::updateValue('PS_PSX_FIREBASE_REFRESH_TOKEN', $response['refresh_token']);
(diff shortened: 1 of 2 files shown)


Quetzacoalt91:
How did you make compliant with multishop feature?

Note the tests did not pass, you need to run php-cs-fixer

Matt75:
> How did you make compliant with multishop feature?

Replace SQL query on configuration table by adding a proper field and use Configuration class to ensure we retrieve date associated to the good shop.
New Configuration key is PS_PSX_FIREBASE_REFRESH_DATE

classes/Api/Firebase/Token.php
```diff
@@ -48,30 +48,48 @@ public function refresh()
...
 
     /**
-     * Check the token validity. The token expire time is set to 3600 seconds.
+     * Check we can request an other token.
+     *
+     * @return bool
+     */
+    public function shouldRefreshToken()
+    {
+        return $this->hasRefreshToken() && $this->isExpired();
```
Quetzacoalt91: What if the token does not exist? `$this->hasRefreshToken()` is false, making the whole line return false too.

Matt75:
Yes, it’s expected :)
We cannot do a call to request an access_token without a refresh_token even if token is expired.
So if false, refresh() is not called
Currently when you are not logged to Psx or you have logout, this field is empty in Configuration but getToken() call refresh() and make call without request data. As a result it throw an exception due to Bad Request, catch by GenericClient
This fix prevent that call
getToken() will return an empty string and will be considered as not logged like expected

classes/Api/Firebase/Token.php
Matt75: Yes, it’s expected :)
We cannot do a call to request an access_token without a refresh_token even if token is expired.
So if false, refresh() is not called
Currently when you are not logged to Psx or you have logout, this field is empty in Configuration but getToken() call refresh() and make call without request data. As a result it throw an exception due to Bad Request, catch by GenericClient
This fix prevent that call
getToken() will return an empty string and will be considered as not logged like expected

Quetzacoalt91:
Thank you @Matt75

PR #219, opened 2019-11-27
Check hooks on Debug page

Matt75:
Replace hook alias by real hook name
Add list of hooks used in debug page and registration status
Need help to improve display 🙃

I found a bug in multistore instance with Module::isRegisteredInHook() was not relevant 🤯so I use Hook::isModuleRegisteredOnHook() instead

Display can be improved
Action to register missing hooks can be added

### DIFF
--- a/ps_checkout.php
+++ b/ps_checkout.php
@@ -41,15 +41,17 @@ if (!defined('_PS_VERSION_')) {
 
 class Ps_checkout extends PaymentModule
 {
-    // hook list used by the module
-    const HOOK_LIST = [
+    /**
+     * @var array Hooks used
+     */
+    public $hooks = [
         'paymentOptions',
-        'paymentReturn',
+        'displayPaymentReturn',
         'actionFrontControllerSetMedia',
         'actionOrderSlipAdd',
-        'orderConfirmation',
+        'displayOrderConfirmation',
         'displayAdminAfterHeader',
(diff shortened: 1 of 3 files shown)


ps_checkout.php
```diff
@@ -44,12 +44,12 @@ class Ps_checkout extends PaymentModule
     // hook list used by the module
     const HOOK_LIST = [
         'paymentOptions',
-        'paymentReturn',
+        'displayPaymentReturn',
```
zalexki: unused hook

ps_checkout.php
```diff
@@ -44,12 +44,12 @@ class Ps_checkout extends PaymentModule
...
         'actionFrontControllerSetMedia',
         'actionOrderSlipAdd',
-        'orderConfirmation',
+        'displayOrderConfirmation',
```
zalexki: does this will still work even tho the method is still called `hookOrderConfirmation` ?

Matt75: Yes, real name and alias are tested on
* Hook::getIdByName()
* Hook::isHookCallableOn()
* Hook::callHookOn()

Moreover on Hook::registerHook() real hook id is used even with alias thanks to Hook::getIdByName()

Better to change hookorderConfirmation to hookDisplayOrderConfirmation indeed

PR #220, opened 2019-11-27
Retry with minimal payload when full payload failed

Matt75:
Retry with minimal payload when full payload failed


Quetzacoalt91:
Thank you @Matt75

PR #221, opened 2019-11-27
Adds logs on api calls

Matt75:
Adds logs on API Calls

### DIFF
--- a/classes/Api/GenericClient.php
+++ b/classes/Api/GenericClient.php
@@ -76,9 +76,20 @@ class GenericClient
     {
         $response = $this->getClient()->post($this->getRoute(), $options);
 
+        /**
+         * @var \Ps_checkout
+         */
+        $module = \Module::getInstanceByName('ps_checkout');
+        $logger = $module->getLogger();
+        $logger->debug('route ' . $this->getRoute());
+        $logger->debug('options ' . var_export($options, true));
+
         $responseHandler = new ResponseApiHandler();
 
-        return $responseHandler->handleResponse($response);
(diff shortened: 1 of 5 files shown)


classes/Api/GenericClient.php
```diff
@@ -76,9 +77,16 @@ protected function post(array $options = [])
     {
         $response = $this->getClient()->post($this->getRoute(), $options);
 
+        $logger = CheckoutLogger::create();
```
zalexki: as we just said let's moove this in the constructor of the module

classes/Factory/CheckoutLogger.php
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+namespace PrestaShop\Module\PrestashopCheckout\Factory;
+
+use Monolog\Handler\RotatingFileHandler;
+use Monolog\Logger;
+
+class CheckoutLogger
+{
+    const MAX_FILES = 7;
```
zalexki: i think @v4lux wanted to have 30 days ?

ps_checkout.php
```diff
@@ -79,6 +80,11 @@ class Ps_checkout extends PaymentModule
     // the module each time to get the version
     const VERSION = '1.2.6';
 
+    /**
+     * @var \Monolog\Logger
+     */
+    public $logger;
```
zalexki: let's be more safe and use a private property and a getter, that way you should be able to do something like this to use it `$logger = \Module::getInstanceByName('ps_checkout')->getLogger();`

PR #224, opened 2019-12-02
EMOD-304 update order states colors

zalexki:


Quetzacoalt91:
These colors will be applied on new module installations. Is it planned to update the current shops?

PR #226, opened 2019-12-02
Enable Release Drafter

Quetzacoalt91:


.github/release-drafter.yml
```diff
@@ -0,0 +1,14 @@
+name-template: v$NEXT_PATCH_VERSION
+tag-template: v$NEXT_PATCH_VERSION
+categories:
+  - title: 🚀 Improvements
+    label: enhancement
+  - title: 🐛 Bug Fixes
+    label: bug
```
v4lux: It should be nice to also add a "feature" category? What do you think? :D

Quetzacoalt91: Added `feature` label from PrestaShop/PrestaShop, then updated the release drafter accordingly

PR #227, opened 2019-12-03
EMOD-229 activate all countries by default

zalexki:


Quetzacoalt91:
To you want to run the same task for merchant upgrading?

zalexki:
no we don't want to

PR #228, opened 2019-12-04
fix 1.7.3 compat

zalexki:


zalexki:
@v4lux this is kinda of huge thing to understand, how this can work on prestashop prior than 1.7.4.0 ?

PR #231, opened 2019-12-05
Run PHPStan with several versions of PrestaShop

Quetzacoalt91:
In order to detect potiental backward compatibility issues with PrestaShop in the future, we run PHPStan with several versions of PS.

This PR contains 2 commits reverting changes, which is expected as proof.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,14 @@ php:
   - 5.6
   - 7.2
 
+env:
+  - PRESTASHOP_VERSION=latest
+
+matrix:
+  include:
+    - php: 7.2
+      env: PRESTASHOP_VERSION=1.7.3.0
+
 cache:
   directories:
     - $HOME/.composer/cache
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -5,6 +5,16 @@ php:
...
 
+env:
+  - PRESTASHOP_VERSION=latest
+
+matrix:
+  include:
+    - php: 7.2
+      env: PRESTASHOP_VERSION=1.7.3.0
+    - php: 7.2
+      env: PRESTASHOP_VERSION=1.7.1.2
```
zalexki: can we add every one of theme like 1.7.0.0 and 1.7.2.0 ?

Quetzacoalt91: We balance between tests, and time needed to run them.

Our objective is to check the backward compatibility with the biggest range of PrestaShop versions. So, checking with the first and the last PrestaShop 1.7.x.x is enough. All other versions in this range have been checked by the core team.

zalexki: hum agree but then we must add 1.7.0.0 because this is where most of the changes happened no ?

zalexki: but now we only have 1.7.0.0 shouldn't we have the latest one too ?

Quetzacoalt91: We do have both:

image

Travis configuration can be confusing sometimes. We already have the 2 builds with the default PS version set as `latest`, on which I added a specific build for PHP 7.2 and PS 1.7.0.0.

zalexki: ok i see seems good then !

Quetzacoalt91:
Regarding the error reported with PHPStan, I decided to not check older version that 1.7.0.3. These were the first 1.7 versions, which we're ready for production use.

Quetzacoalt91:
Thanks for review

PR #235, opened 2019-12-16
Fix default card icon on theme without material-icon

Matt75:
Fix default card icon on theme without material-icon instead of display `credit_card`on card number field


Quetzacoalt91:
Thank you @Matt75

PR #237, opened 2019-12-17
update cgu links

zalexki:


classes/Presenter/Store/Modules/ContextModule.php
```diff
@@ -120,7 +120,25 @@ private function getReadme()
...
+                break;
+            case 'es':
+                return 'https://www.prestashop.com/es/prestashop-checkout-condiciones-generales-uso';
+                break;
+            case 'it':
+                return 'https://www.prestashop.com/it/prestashop-checkout-condizioni-generali-utilizzo';
+                break;
+            case 'en':
+                return 'https://www.prestashop.com/en/prestashop-checkout-general-terms-use';
+                break;
```
Quetzacoalt91: English is already handled in the `default` case, and can be removed.

Quetzacoalt91:
Thank you @zalexki

PR #240, opened 2019-12-27
Feature/express checkout

v4lux:


controllers/front/ExpressCheckout.php
```diff
@@ -0,0 +1,234 @@
...
+    private function createAddress($shipping)
+    {
+        // check if country is available for delivery
+        $paypalIsoCode = $shipping->address->country_code;
+        $psIsoCode = (new PaypalCountryCodeMatrice())->getPrestashopIsoCode($paypalIsoCode);
+        $idCountry = \Country::getByIso($psIsoCode);
+
+        $country = new \Country($idCountry);
+
+        if ('0' == (string) $country->active) {
```
Quetzacoalt91: If you take time to cast one of the variable, you can make a strict comparison with `===`.

```php
        if ('0' === (string) $country->active) {
```

But wouldn't it cleaner to check integers instead?

```php
        if (0 === (int) $country->active) {
```

controllers/front/ExpressCheckout.php
```diff
@@ -0,0 +1,234 @@
...
+        // Retry with minimal payload when full payload failed
+        if (substr((string) $response['httpCode'], 0, 1) === '4') {
+            $builder->buildMinimalPayload();
+            $payload = $builder->presentPayload()->getJson();
+            $response = (new Order($this->context->link))->create($payload);
+        }
+
+        if (false === $response['status']) {
+            return false;
+        }
```
Quetzacoalt91: I fsomething went wrong, should we set another HTTP response code than 200?

views/js/api-paypal.js
```diff
@@ -215,12 +220,12 @@ function hideDefaultPaymentButtonIfPaypalIsChecked() {
...
 
-  let paypalOptions = document.getElementsByName('payment-option');
+  const paypalOptions = document.getElementsByName('payment-option');
 
   for (let i = 0; i < paypalOptions.length; i++) {
-    let item = paypalOptions[i];
+    const item = paypalOptions[i];
 
-    item.addEventListener('click', function() {
+    item.addEventListener('click', () => {
```
Quetzacoalt91: IE11 isn't compatible with this method definition

views/js/expressCheckout.js
```diff
@@ -0,0 +1,180 @@
...
+    });
+  }
+
+  if (displayMode === 'product') {
+    const addToCartButton = document.getElementsByClassName('product-add-to-cart');
+    addToCartButton[0].parentNode.insertBefore(document.getElementById('pscheckout-express-checkout'), addToCartButton[0].nextSibling);
+  }
+
+  // wait paypal sdk to be fully loaded
+  const interval = setInterval(() => {
```
Quetzacoalt91: prefer using `function()` instead of `() => {` for compatibility with IE11

views/js/expressCheckout.js
```diff
@@ -0,0 +1,180 @@
...
+        redirect(order, controllerLink);
+      });
+    },
+  }).render('#paypal-button-container');
+}
+
+/**
+ * Ajax request to create and retrieve paypal order
+ */
+async function createPaypalOrder() {
```
Quetzacoalt91: To be tested on IE11. Looks like `async` isn't compatible with this browser https://caniuse.com/#feat=async-functions

Quetzacoalt91: Compiling these JS scripts for compatibility with old browser may help

views/js/expressCheckout.js
```diff
@@ -0,0 +1,180 @@
...
+  const controllerLink = expressCheckoutController.replace(/\amp;/g, '');
+
+  return new Promise((resolve, reject) => {
```
Quetzacoalt91: Same comment here.

views/templates/front/expressCheckout.tpl
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @author    PrestaShop SA <contact@prestashop.com>
+ * @copyright 2007-2019 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ *}
+
+<script type='text/javascript' src='{$jsExpressCheckoutPath|escape:'javascript':'UTF-8'}'></script>
+
+<link rel="preload" href="https://www.paypal.com/sdk/js?components=buttons&amp;client-id={$paypalClientId|escape:'htmlall':'UTF-8'}&amp;merchant-id={$merchantId|escape:'htmlall':'UTF-8'}&amp;intent={$intent|escape:'htmlall':'UTF-8'}&amp;currency={$currencyIsoCode|escape:'htmlall':'UTF-8'}&amp;commit=false&amp;disable-funding=credit,card" as="script">
```
Quetzacoalt91: Did you really need to write `&amp;` instead of `&`?

views/templates/front/expressCheckout.tpl
```diff
@@ -0,0 +1,70 @@
...
+
+<div id="pscheckout-express-checkout" style="display:none;">
+  {if $displayMode eq 'cart'}
+  <div style="margin-top:15px;margin-bottom:15px">{l s='or' mod='ps_checkout'}</div>
```
Quetzacoalt91: These style definition should be in a css file, otherwise other themes can't modify them if needed.

views/templates/front/expressCheckout.tpl
```diff
@@ -0,0 +1,70 @@
...
+  {/if}
+
+  {if $displayMode eq 'checkout'}
+  <div style="margin-top:15px;margin-bottom:15px">
```
Quetzacoalt91: Same comment about style definition

PR #239, opened 2019-12-23
Add docker nodejs

zalexki:

PR #241, opened 2019-12-30
Reporting lot1

zalexki:
/!\ Need to wait #239 to be merged.

waiting information from @v4lux about commission on orders

Also i think the missing front styling on the table should be handled by the presta ui kit.

This PR contain also some code for the next lot.

### DIFF
--- a/_dev/src/components/block/reporting.vue
+++ b/_dev/src/components/block/reporting.vue
@@ -26,15 +26,16 @@
           <a href="">{{ $t('block.reporting.subtitleLinkLabel') }}</a>
         </p>
 
+        <!-- Orders table -->
         <b-table
           show-empty
           hover
           :items="orders"
-          :fields="fields"
-          :filter="filter"
-          :current-page="currentPage"
-          :per-page="perPage"
-          @filtered="onFiltered"
+          :fields="orderFields"
+          :filter="orderFilter"
+          :current-page="orderCurrentPage"
+          :per-page="orderPerPage"
+          @filtered="onOrderFiltered"
         >
           <template v-slot:cell(user)="data">
             <a :href="`${data.item.userProfileLink}`">
@@ -54,13 +55,32 @@
         </b-table>
 
         <b-pagination
-          v-model="currentPage"
-          :total-rows="totalRows"
-          :per-page="perPage"
+          v-model="orderCurrentPage"
+          :total-rows="orderTotalRows"
+          :per-page="orderPerPage"
           align="fill"
           size="sm"
           class="my-0"
         ></b-pagination>
+
+        <!-- Transactions table -->
+        <b-table
+          show-empty
+          hover
+          :items="transactions"
+          :fields="transactionFields"
+          :filter="transactionFilter"
+          :current-page="transactionCurrentPage"
+          :per-page="transactionPerPage"
+          @filtered="onTransactionFiltered"
+        >
+          <template v-slot:cell(user)="data">
+            <a :href="`${data.item.userProfileLink}`">
+              {{ data.item.username }}
+            </a>
+          </template>
+        </b-table>
+
       </div>
     </b-card-body>
   </b-card>
@@ -72,9 +92,13 @@
   export default {
     name: 'Reporting',
     methods: {
-      onFiltered(filteredItems) {
-        this.totalRows = filteredItems.length;
-        this.currentPage = 1;
+      onOrderFiltered(filteredItems) {
+        this.orderTotalRows = filteredItems.length;
+        this.orderCurrentPage = 1;
+      },
+      onTransactionFiltered(filteredItems) {
+        this.transactionTotalRows = filteredItems.length;
+        this.transactionCurrentPage = 1;
       },
       getReportingDatas() {
         ajax({
@@ -82,7 +106,9 @@
           action: 'GetReportingDatas',
         }).then((response) => {
           this.orders = response.orders;
-          this.totalRows = this.orders.length;
+          this.transactions = response.transactions;
+          this.orderTotalRows = this.orders.length;
+          this.transactionTotalRows = this.transactions.length;
         });
       },
       setBadgeColor(color) {
@@ -93,10 +119,7 @@
     },
     data() {
       return {
-        filter: null,
-        filterOn: [],
-        orders: [],
-        fields: [
+        orderFields: [
           {key: 'date_add', label: 'Date', sortable: true},
           {key: 'id_order', label: 'Order ID', sortable: true},
           {key: 'user', label: 'Customer', sortable: true},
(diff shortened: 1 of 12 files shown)


zalexki:
seems ok on 1.6 let's review and merge !

classes/Presenter/Order/OrderPendingPresenter.php
```diff
@@ -0,0 +1,114 @@
+<?php
+/**
+* 2007-2019 PrestaShop
```
apacios: We should update this to 2020

classes/Presenter/Order/OrderPendingPresenter.php
```diff
@@ -0,0 +1,114 @@
...
+     * @param array $idStates
+     *
+     * @return mixed
+     */
+    private function getPendingOrders($idStates)
+    {
+        $sql = 'SELECT id_order, current_state, total_paid, date_add, id_customer
+            FROM `' . _DB_PREFIX_ . 'orders` o
+            WHERE o.module = "ps_checkout"
+            AND current_state IN (' . implode(',', array_keys($idStates)) . ')
```
apacios: To give the best example we should use `array_map` with the `implode` to force cast into `int`

classes/Presenter/Transaction/TransactionPresenter.php
```diff
@@ -0,0 +1,106 @@
...
+            LIMIT 1000
+        ';
+
+        return \Db::getInstance()->executeS($sql);
+    }
+
+    private function getOrderIDByOrderReference($reference)
+    {
+        $sql = 'SELECT id_order FROM `' . _DB_PREFIX_ . 'orders` o
+            WHERE reference = "' . $reference . '"
```
apacios: To give the best example we may use pSQL here

classes/Presenter/Transaction/TransactionPresenter.php
```diff
@@ -0,0 +1,106 @@
...
+     * getUserInfos
+     *
+     * @param string $orderReference
+     *
+     * @return string
+     */
+    private function getUserInfos($orderReference)
+    {
+        $link = new LinkAdapter();
+        $sql = 'SELECT id_customer FROM `' . _DB_PREFIX_ . 'orders` WHERE reference = "' . $orderReference . '"';
```
apacios: To give the best example we may use pSQL here

_dev/src/components/block/reporting.vue
```diff
@@ -0,0 +1,244 @@
+<!--**
+ * 2007-2019 PrestaShop and Contributors
```
apacios: We should update this to 2020

classes/Presenter/Order/OrderPendingPresenter.php
```diff
@@ -0,0 +1,114 @@
...
+* to license@prestashop.com so we can send you a copy immediately.
+*
+* DISCLAIMER
+*
+* Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+* versions in the future. If you wish to customize PrestaShop for your
+* needs please refer to http://www.prestashop.com for more information.
+*
+*  @author    PrestaShop SA <contact@prestashop.com>
+*  @copyright 2007-2019 PrestaShop SA
```
apacios: We should update this to 2020.
The license seems not to be the same in all files ?

classes/Presenter/Order/index.php
```diff
@@ -0,0 +1,28 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
```
apacios: We should update this to 2020

classes/Presenter/Transaction/TransactionPresenter.php
```diff
@@ -0,0 +1,106 @@
+<?php
+/**
+* 2007-2019 PrestaShop
```
apacios: We should update this to 2020

classes/Presenter/Transaction/index.php
```diff
@@ -0,0 +1,28 @@
+<?php
+/**
+ * 2007-2019 PrestaShop and Contributors
```
apacios: We should update this to 2020

classes/Presenter/Order/OrderPendingPresenter.php
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * getUsername
+     *
+     * @param string|int $userID
+     *
+     * @return string
+     */
+    private function getUsername($userID)
+    {
+        $sql = 'SELECT firstname,lastname FROM `' . _DB_PREFIX_ . 'customer` o WHERE id_customer = ' . $userID;
```
v4lux: It should be better to instantiate a customer object ? It will be more easier to read and it will avoid to use substr too.

classes/Presenter/Transaction/TransactionPresenter.php
v4lux: You should take a look to the method getByReference() (available since ps 1.5.0.14) in the Order class. You will be able to retrieve the customer id.

classes/Presenter/Transaction/TransactionPresenter.php
```diff
@@ -0,0 +1,106 @@
...
+     *
+     * @return string
+     */
+    private function getUserInfos($orderReference)
+    {
+        $link = new LinkAdapter();
+        $sql = 'SELECT id_customer FROM `' . _DB_PREFIX_ . 'orders` WHERE reference = "' . $orderReference . '"';
+        $userID = \Db::getInstance()->getRow($sql);
+
+        $sql = 'SELECT firstname,lastname FROM `' . _DB_PREFIX_ . 'customer` WHERE id_customer = ' . (int) $userID['id_customer'];
```
v4lux: Instantiate customer ? I don't know if a direct query to the Database is faster than instantiate an object but i Think it is more readable.

controllers/admin/AdminAjaxPrestashopCheckoutController.php
```diff
@@ -218,6 +220,24 @@ public function ajaxProcessGetOnboardingLink()
         );
     }
 
+    /**
+     * AJAX: Retrieve Reporting informations
+     */
+    public function ajaxProcessGetReportingDatas()
+    {
+        $sql = 'SELECT COUNT(id_order_payment) FROM `' . _DB_PREFIX_ . 'order_payment` o WHERE o.payment_method = "Prestashop Checkout"';
```
v4lux: Use prestashop collection instead ?

classes/Presenter/Transaction/TransactionPresenter.php
v4lux: getByReference() from Order class ?

zalexki: thx for the tip, i'll stay on the custom query adding `pSQL` to it

classes/Presenter/Transaction/TransactionPresenter.php
zalexki: thx but same i'll just add pSQL

controllers/admin/AdminAjaxPrestashopCheckoutController.php
zalexki: no sql count is better

classes/Presenter/Order/OrderPendingPresenter.php
zalexki: i don't see why ? 🤔

classes/Presenter/Transaction/TransactionPresenter.php
zalexki: in order to instantiate the object you need the database infos, and like doctrine the prestashop object model will load every props of an object so it might be slightly more readable but at some perf cost that IMO is not enough since here we have a very small query.
@apacios @Matt75 wdyt ?

classes/Presenter/Order/OrderPendingPresenter.php
apacios: I don't know if it's usefull but we also can do : 
`return \db::getInstance()->getValue('SELECT CONCAT(SUBSTR(firstname, 1, 1), ". ", lastname) FROM `' . _DB_PREFIX_ . 'customer` o WHERE id_customer = ' . $userID);`

classes/Presenter/Transaction/TransactionPresenter.php
apacios: As it is a simple query it seems good to me like this. It's readable and it's a realy quick query.

classes/Presenter/Order/OrderPendingPresenter.php
zalexki: ye ok thx for the tip, not really better or something just different in our case

v4lux: Forget it, you will still need to do a substr to get the first letter.

classes/Presenter/Order/OrderPendingPresenter.php
```diff
@@ -101,6 +95,7 @@ private function getUsername($userID)
      */
     private function getPendingOrders($idStates)
     {
+        $idStates = array_map([$this, 'castToInt'], $idStates);
```
v4lux: Try to use intval() instead ?

PR #258, opened 2020-01-13
Express checkout on checkout page on ps < 1.7.6

v4lux:


ps_checkout.php
```diff
@@ -212,11 +213,50 @@ public function uninstallTabs()
...
+        }
+
+        if (true === $this->isPaymentStep()) {
+            return false;
+        }
+
+        $expressCheckout = new ExpressCheckout($this, $this->context);
+        $expressCheckout->setDisplayMode(ExpressCheckout::CHECKOUT_MODE);
+
+        return $expressCheckout->render();
```
Quetzacoalt91: Instead of copy pasting code, please create a private method to call in your two hooks.

PR #261, opened 2020-01-15
Fix json encoding issue that occurs in some cases on express checkout…

v4lux:
… (product page)

### DIFF
--- a/views/js/initExpressCheckout.js
+++ b/views/js/initExpressCheckout.js
@@ -109,8 +109,15 @@ async function createPaypalOrder() {
     form.append('action', 'CreatePaypalOrder');
 
     if (displayMode === 'product') {
-      const productDetails = document.getElementById('product-details').dataset.product;
-      form.append('product', productDetails);
+      const productDetails = JSON.parse(document.getElementById('product-details').dataset.product);
+      const product = {
+        'id_product': productDetails.id_product,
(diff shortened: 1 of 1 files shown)


views/js/initExpressCheckout.js
```diff
@@ -109,8 +109,15 @@ async function createPaypalOrder() {
...
-      form.append('product', productDetails);
+      const productDetails = JSON.parse(document.getElementById('product-details').dataset.product);
+      const product = {
+        'id_product': productDetails.id_product,
+        'id_product_attribute': productDetails.id_product_attribute,
+        'id_customization': productDetails.id_customization,
+        'quantity_wanted': productDetails.quantity_wanted,
+      }
+
+      form.append('product', JSON.stringify(product));
```
Quetzacoalt91: Do we absolutely have to stringify `product` by ourselves? Or sending the object to the FormData would also automatically trigger it?

v4lux: Just tested, doesn't work. If you want I can add 4 key to the formData instead of making an object? It avoid to encode on the js side and decoding it after in the php side (controller) ?

Quetzacoalt91: I'm good with the current changes, no need to split your objects in 4 keys. :)

PR #262, opened 2020-01-17
Avoid to throw an exception when we cannot update the status of the p…

v4lux:
…aypal account and create a waiting status


Quetzacoalt91:
Does this PR brings an improvement or a bug fix?

_dev/src/store/modules/paypal/actions.js
```diff
@@ -43,4 +43,13 @@ export default {
       return Promise.resolve();
     });
   },
+  updatePaypalStatus({commit, getters}) {
```
Quetzacoalt91: By using `update`, I understand that you're trying to send data to store to a controller. `refreshPaypalStatus` seems more appropriate.

classes/Entity/PaypalAccount.php
```diff
@@ -71,10 +71,6 @@ class PaypalAccount
 
     public function __construct($merchantId = null, $email = null, $emailIsVerified = null, $paypalPaymentStatus = null, $cardPaymentStatus = null)
     {
-        if (empty($merchantId)) {
-            throw new \PrestaShopException('merchantId cannot be empty');
-        }
```
Quetzacoalt91: `$merchantId` is now an optional parameter?

_dev/src/App.vue
```diff
@@ -72,6 +72,18 @@
...
+    data() {
+      return {
+        paypalStatusUpdater: null,
+      };
+    },
+    methods: {
+      updater() {
+        this.paypalStatusUpdater = setInterval(() => {
+          this.$store.dispatch('updatePaypalStatus');
+        }, 5000);
```
Quetzacoalt91: This value seems low, I guess we can double it to 10s.

_dev/src/store/modules/paypal/actions.js
v4lux: Renamed to refreshPaypalStatus() for a better understanding.

_dev/src/App.vue
v4lux: Changed to 10 seconds.

PR #268, opened 2020-01-24
Fix wrong column type for pscheckout_order_matrice table

Matt75:
Change wrong column type for id_order_prestashop  in pscheckout_order_matrice table

### DIFF
--- a/upgrade/upgrade-1.2.11.php
+++ b/upgrade/upgrade-1.2.11.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * 2007-2020 PrestaShop and Contributors
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Academic Free License 3.0 (AFL-3.0)
+ * that is bundled with this package in the file LICENSE.txt.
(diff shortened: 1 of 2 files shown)


upgrade/upgrade-1.2.11.php
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Update main function for module Version 1.2.11
+ *
+ * @param Module $module
+ *
+ * @return bool
+ */
+function upgrade_module_1_2_11($module)
+{
+    return \Db::getInstance()->execute('ALTER TABLE `' . _DB_PREFIX_ . OrderMatrice::TABLE_ORDER_MATRICE . '` CHANGE `id_order_prestashop` `id_order_prestashop` INT(10) UNSIGNED NOT NULL;');
```
Quetzacoalt91: `OrderMatrice::TABLE_ORDER_MATRICE` does not exist, it should be `TableManager::TABLE_ORDER_MATRICE` (namespace to be updated)

Matt75: Done

PR #270, opened 2020-01-29
fix make command build-zip-prod

zalexki:


Makefile
```diff
@@ -38,7 +38,11 @@ build-zip:
 	rm -rf $(PWD)/../buildedZIP/tests
 	rm -rf $(PWD)/../buildedZIP/docker-compose.yml
 	rm -rf $(PWD)/../buildedZIP/Makefile
-	zip ps_checkout.zip $(PWD)/../buildedZIP
+	mkdir ps_checkout
+	mv -v $(PWD)/../buildedZIP/* ps_checkout
+	cp $(PWD)/../buildedZIP/.env ps_checkout/.env
+	zip -r ps_checkout.zip ps_checkout
```
Quetzacoalt91: It seems you're moving again the content in a `ps_checkout` folder to allow the module content to be in a subdirectory inside the zip. To prevent issues like the missing file `.env`, work directly with `ps_checkout/` instead of `$(PWD)/../buildedZIP`. This will go faster and you won't need to `mv` & `cp .env`.

Also the code to generate archives is duplicated, there is one here and another one in Travis. We should avoid this and reuse when possible. BTW @matks started a POC which could help all module to build these artifacts easily: https://github.com/matks/ps-modules-builder

zalexki: i resolve this discussion because i pushed a commit that now use the /tmp folder and handle a little more gracefully this zip creation

Quetzacoalt91: Agreed

PR #278, opened 2020-01-31
Add rate module component

apacios:


apacios:
Do not merge until we confirm that it works

Matt75:
Free module cannot be rated on PrestaShop Addons.
Customer account is mandatory with a download from PrestaShop Checkout product page.
This is not possible for modules downloaded from Module Catalog in PrestaShop back office.

PR #280, opened 2020-02-04
Handle PayPal capabilities status for hosted fields

Matt75:
Some account has PayPal product PPCP_CUSTOM (hosted fields) to SUBSCRIBED
But account has PayPal capabilities CUSTOM_CARD_PROCESSING to SUSPENDED or REVOKED
We need to use it to display right status in BO and no display hosted fields in FO if account not eligible
https://developer.paypal.com/docs/api/partner-referrals/v1/#definition-account_capability


v4lux:
This PR is deprecated in favor of #337.

PR #281, opened 2020-02-05
Multishop compatibility

Matt75:
Force one configuration per shop
Access to module configuration page is not allowed if Context is All Shops or Group
OrderState not manage multishop that's why I use globalvalue

### DIFF
--- a/_dev/src/App.vue
+++ b/_dev/src/App.vue
@@ -41,13 +41,16 @@
     <div class="pt-5" />
     <div class="pt-3" />
 
-    <div class="container">
+    <div
+      class="container"
+      v-if="isShopContext"
+    >
       <RoundingBanner />
     </div>
 
     <div
       class="container"
-      v-if="paymentMode === 'SANDBOX'"
+      v-if="isTestMode"
     >
       <b-alert
         variant="warning"
@@ -56,7 +59,37 @@
         <p>{{ $t('general.testModeOn') }}</p>
       </b-alert>
     </div>
-    <router-view />
+
+    <div
+      class="container"
+      v-if="!isShopContext"
+    >
+      <b-alert
+        variant="warning"
+        show
+      >
+        <h2>{{ $t('general.multiShop.title') }}</h2>
+        <p>{{ $t('general.multiShop.subtitle') }}</p>
+        <p>{{ $t('general.multiShop.chooseOne') }}</p>
+        <b-list-group
+          v-for="group in shopsTree"
+          :key="group.id"
+          class="mt-3 mb-3 col-4"
+        >
+          <p class="text-muted">{{ $t('general.multiShop.group') }} {{ group.name }}</p>
+          <b-list-group-item
+            v-for="shop in group.shops"
+            :key="shop.id"
+            :href="shop.url"
+          >
+            {{ $t('general.multiShop.configure') }} <b>{{ shop.name }}</b>
+          </b-list-group-item>
+        </b-list-group>
+        <p>{{ $t('general.multiShop.tips') }}</p>
+      </b-alert>
+    </div>
+
+    <router-view v-if="isShopContext" />
   </div>
 </template>
 
@@ -91,11 +124,18 @@
       onboardingFirebaseIsCompleted() {
         return this.$store.state.firebase.onboardingCompleted;
       },
-      paymentMode() {
-        return this.$store.state.configuration.paymentMode;
-      },
       accountIslinked() {
         return this.$store.state.paypal.accountIslinked;
+      },
+      isShopContext() {
+        return this.$store.state.context.isShopContext;
+      },
+      isTestMode() {
+        return this.$store.state.configuration.paymentMode === 'SANDBOX'
+          && this.$store.state.context.isShopContext;
+      },
+      shopsTree() {
+        return this.$store.state.context.shopsTree;
       },
     },
     watch: {
--- a/classes/Api/Firebase/Token.php
+++ b/classes/Api/Firebase/Token.php
@@ -41,14 +41,37 @@ class Token extends FirebaseClient
         $response = $this->post([
             'json' => [
                 'grant_type' => 'refresh_token',
-                'refresh_token' => \Configuration::get('PS_PSX_FIREBASE_REFRESH_TOKEN'),
+                'refresh_token' => \Configuration::get(
+                    'PS_PSX_FIREBASE_REFRESH_TOKEN',
+                    null,
+                    null,
+                    (int) \Context::getContext()->shop->id
+                ),
             ],
         ]);
 
         if (true === $response['status']) {
-            \Configuration::updateValue('PS_PSX_FIREBASE_ID_TOKEN', $response['body']['id_token']);
-            \Configuration::updateValue('PS_PSX_FIREBASE_REFRESH_TOKEN', $response['body']['refresh_token']);
-            \Configuration::updateValue('PS_PSX_FIREBASE_REFRESH_DATE', date('Y-m-d H:i:s'));
+            \Configuration::updateValue(
+                'PS_PSX_FIREBASE_ID_TOKEN',
+                $response['body']['id_token'],
+                false,
+                null,
+                (int) \Context::getContext()->shop->id
+            );
+            \Configuration::updateValue(
+                'PS_PSX_FIREBASE_REFRESH_TOKEN',
+                $response['body']['refresh_token'],
+                false,
+                null,
+                (int) \Context::getContext()->shop->id
+            );
+            \Configuration::updateValue(
+                'PS_PSX_FIREBASE_REFRESH_DATE',
+                date('Y-m-d H:i:s'),
+                false,
+                null,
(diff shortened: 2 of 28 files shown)


classes/Api/Firebase/Token.php
```diff
@@ -41,14 +41,37 @@ public function refresh()
...
+                false,
+                null,
+                (int) \Context::getContext()->shop->id
+            );
+            \Configuration::updateValue(
+                'PS_PSX_FIREBASE_REFRESH_DATE',
+                date('Y-m-d H:i:s'),
+                false,
+                null,
+                (int) \Context::getContext()->shop->id
```
Quetzacoalt91: What do you think about having an adapter for `Configuration` in `classes/Adapter`, allowing you to reduce the number of parameters in all other places of the project?

Matt75: We don't have container yet to reuse instance for this type of Service.
Each time we need to use it, we have to instantiate it and if we use namespace, we have to add complete namespace in all place we use it in main module class and upgrade files...
Currently others adapters are instantiated more than one time at each execution and i'm not satisfied of that too.
I know it's boring to give each parameters to Configuration method but I don't find more faster to call an Adapter to do that with full namespace in main class.
I think we have to handle this kind of needs we a huge reflexion about architecture we need to use.

Quetzacoalt91: I totally agree, having an equivalent of a container providing dependencies to each class of this project would be helpful.
I suggest to work on this idea because even if you did everything well in this PR, we can't be sure that the next person modyfing this project will think about using all the parameters.

Matt75: This feature is in top priority, so I suggest to handle architecture work on another pull request.
It's a very interesting topic and need huge work, maybe to be coupled with work on exceptions

Quetzacoalt91: Indeed this would require too many changes and work, this has to be done in another PR. :)

classes/Translations/Translations.php
```diff
@@ -56,6 +56,13 @@ public function getTranslations()
             'general' => [
                 'save' => $this->module->l('Save', 'translations'),
                 'testModeOn' => $this->module->l('Test mode is turned on', 'translations'),
+                'multiShop' => [
+                    'title' => $this->module->l('Multistore detected', 'translations'),
+                    'subtitle' => $this->module->l('You must configure your shops one by one for this service, but you will be able to use the same accounts.', 'translations'),
+                    'chooseOne' => $this->module->l('Please select the first shop to configure from the list below :', 'translations'),
+                    'group' => 'Group :',
```
Quetzacoalt91: There's no space before a colon in English:

```suggestion
                    'group' => 'Group:',
```

classes/Translations/Translations.php
```diff
@@ -56,6 +56,13 @@ public function getTranslations()
             'general' => [
                 'save' => $this->module->l('Save', 'translations'),
                 'testModeOn' => $this->module->l('Test mode is turned on', 'translations'),
+                'multiShop' => [
+                    'title' => $this->module->l('Multistore detected', 'translations'),
+                    'subtitle' => $this->module->l('You must configure your shops one by one for this service, but you will be able to use the same accounts.', 'translations'),
```
Quetzacoalt91: @LouiseBonnard, may we ask for your help on this PR?

```suggestion
                    'subtitle' => $this->module->l('Each shop must be configured separately, even if you configure the same account on all of them.', 'translations'),
```

classes/Translations/Translations.php
```diff
@@ -56,6 +56,13 @@ public function getTranslations()
...
+                    'chooseOne' => $this->module->l('Please select the first shop to configure from the list below :', 'translations'),
+                    'group' => 'Group :',
+                    'tips' => $this->module->l('After you are done with the first shop, you can configure the others, by selecting them one by one in the shop selector, in the horizontal menu.', 'translations'),
```
Quetzacoalt91: Another suggestion, to be reviewed by @LouiseBonnard & @Darmona 

```suggestion
                    'tips' => $this->module->l('When you are done with a shop, another one can be selected by using the shop selector at the top of the page.', 'translations'),
```

_dev/src/App.vue
```diff
@@ -56,7 +56,37 @@
...
+          :key="group.id"
+          class="mt-3 mb-3 col-4"
+        >
+          <p class="text-muted">{{ $t('general.multiShop.group') }} {{ group.name }}</p>
+          <b-list-group-item
+            v-for="shop in group.shops"
+            :key="shop.id"
+            :href="shop.url"
+          >
+            {{ shop.name }}
```
Quetzacoalt91: If you plan to display buttons, do it the same way as the message above it with `<b-button>` so everything remain coherent:

image

Having a call to action would help too, something like "Configure xxx" or "Switch on xxx"

Matt75: Based on screen made by UX
It's not buttons but list elements, idea is to have a list of shops displayed by group like the native shop selector in header

classes/Presenter/Store/Modules/ContextModule.php
```diff
@@ -75,6 +82,57 @@ public function present()
...
+    }
+
+    /**
+     * @return array
+     */
+    private function getShopsTree()
+    {
+        $shopList = [];
+
+        if (false === $this->isShopContext()) {
```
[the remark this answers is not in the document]
Matt75: Done

classes/Translations/Translations.php
Matt75: Done

controllers/front/PaymentCard16.php
```diff
@@ -68,7 +68,12 @@ public function initContent()
             'clientToken' => $paypalOrder['body']['client_token'],
             'paypalOrderId' => $paypalOrder['body']['id'],
             'validateOrderLinkByCard' => $module->getValidateOrderLink($paypalOrder['body']['id'], 'card'),
-            'intent' => strtolower(\Configuration::get('PS_CHECKOUT_INTENT')),
+            'intent' => strtolower(Configuration::get(
```
zalexki: we don't have strtolower there https://github.com/PrestaShopCorp/ps_checkout/pull/281/files#diff-3d415b068dfb8918e1bdee4e51078c7eR127
if we can avoid it everywhere it would be better IMO

Matt75: Done

Quetzacoalt91:
Wait before merging as I'm waiting for other reviews.

Darmona:
PM approved comportment, on Matt laptop.
I didn't test if webhook distribution goes well, but @Matt75  have you tested?

_dev/src/App.vue
Quetzacoalt91: Following Nadia's feedback, roundings may be not be applied to the proper context.

Shall we display the alerts about rounding & the Sandbox module only whenthe context is on a specific shop?

Quetzacoalt91: Following Nadia's feedback, roundings may be not be applied to the proper context.

Shall we display the alerts about rounding & the Sandbox module only when the context is on a specific shop?

Darmona: Oh right, as long as you can't display PrestaShop Checkout in MultiShop, make no sense to display roudings banner or sandbox mode banner in multishop neither.

For the call to action, not a bad idea.
Could we display 
"Group: Default
Configure PrestaShop
Configure Doge
Configure Default Shop"

?

Darmona: Oh right, as long as you can't display PrestaShop Checkout in MultiShop, make no sense to display roudings banner or sandbox mode banner in multishop neither.

For the call to action, not a bad idea.
Could we display 
"Group: Default
Configure PrestaShop
Configure Doge
Configure Default Shop"

?

NadiaCamard:
When several alerts are displayed, the "multishop" message should appear first (at the top).

What happens when I click on the "Roundings" message, in that case? Which shop the change will be applied on?

classes/Translations/Translations.php
LouiseBonnard: Looks clear to me!

classes/Translations/Translations.php
LouiseBonnard:
```suggestion
                    'tips' => $this->module->l('Once you are done with the first shop, you can configure the others: select them one by one with the shop selector, in the horizontal menu.', 'translations'),
```

classes/Translations/Translations.php
Matt75: Done, thanks for help

classes/Translations/Translations.php
Matt75: Done, thanks for help

_dev/src/App.vue
Matt75: I added Configure before shop name

Matt75:
Update: when alert about multistore is displayed, others are not displayed because it was not relevant.
Webhook distribution is ok, each shop has is own url for webhook.

PR #282, opened 2020-02-05
Adding GitHub actions covering Travis steps

Quetzacoalt91:
We complete the test & build process with GitHub actions, replacing in the future the Travis jobs and being used as a reference for other modules

ToDo:
* [x] Squash all commits to avoid the same title in 20 commits


.github/workflows/js.yml
```diff
@@ -1,14 +1,28 @@
 name: JS tests
 on: [push, pull_request]
 jobs:
-  js-linter:
-    name: JS linter
```
Quetzacoalt91: The previous name was okay (BTW please start your command names with a capital letter)

.github/workflows/js.yml
```diff
@@ -1,14 +1,28 @@
...
-      - uses: actions/setup-node@v1
+      - name: lint js dependencies
+        uses: PoulainMaxime/github-action-lint-js/12@master
         with:
-          node-version: '13.x'
-      - run: yarn --cwd _dev/ install
-      - run: yarn --cwd _dev/ lint
+          cmd: yarn
+          path: ./_dev
+      - name: build js dependencies
```
Quetzacoalt91: Actually, from this step we could create a new workflow (`build.yml`) responsible in downloading / compiling dependencies, cleaning then build / uploading the artifact.

If possible, make sure to start it **after** the tests have been run successfully.

.github/workflows/js.yml
```diff
@@ -1,14 +1,28 @@
...
+        uses: PrestaShopCorp/github-action-build-js/12@dev
+        with:
+          cmd: yarn
+          path: ./_dev
+      - name: Install composer dependencies
+        run: composer install --no-dev -o
+      - name: clean dependencies
```
Quetzacoalt91:
```suggestion
      - name: Clean-up project
```

Quetzacoalt91:
image

The artifact seems to be good!
Please squash your commits then we're good to go :raised_hands:

.github/workflows/build.yml
```diff
@@ -0,0 +1,23 @@
...
+        uses: actions/checkout@v2.0.0
+      - name: Build JS dependencies
+        uses: PrestaShopCorp/github-action-build-js/12@dev
+        with:
+          cmd: yarn
+          path: ./_dev
+      - name: Install composer dependencies
+        run: composer install --no-dev -o
+      - name: Clean-up project
+        uses: PrestaShopCorp/github-action-clean-before-deploy@dev
```
apacios: Before accepting your PR, can't we release github-action-clean-before-deploy to set a specific version there ?
Like `PrestaShopCorp/github-action-clean-before-deploy@v1` ?

PoulainMaxime: Is there really a use for doing it as a version and not using a branch?
Currently I use the dev branch because the github action is currently in code review for merge in master.

zalexki: i agree with @apacios, this is a dependency like any others, so if you update a branch you will possibly break every repository depending on it :/

PoulainMaxime: no problem when github actions will be on master I'll make a tag with a specific version

PoulainMaxime: The github actions are on the master branch and the tags have been created.

.github/workflows/build.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2.0.0
+      - name: Build JS dependencies
+        uses: PrestaShopCorp/github-action-build-js/12@v1.0
+        with:
+          cmd: yarn
+          path: ./_dev
+      - name: Install composer dependencies
+        run: composer install --no-dev -o
```
zalexki: do we know with which php version do we build this ?

.github/workflows/build.yml
```diff
@@ -0,0 +1,23 @@
+name: Deploy
+on: [push, pull_request]
```
zalexki: let's clean this using only pull request and release events and add the artifact to the releases published

.github/workflows/js.yml
```diff
@@ -7,8 +7,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2.0.0
-      - uses: actions/setup-node@v1
+      - name: lint js dependencies
```
zalexki: we don't lint the dependencies but our own code

.github/workflows/build.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  deploy:
+    name: build dependencies & create artifact
```
zalexki: we build the project, the dependencies are juste installed

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+          path: ../
+      - name: Publish to GitHub Release
+        if: github.event_name == 'release' && github.event.action == 'published'
+        uses: actions/upload-release-asset@v1.0.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.release_info.outputs.upload_url }}
+          asset_path: ./build/ps_checkout.zip
+          asset_name: ps_checkout.zip
```
Quetzacoalt91: It would be great to find a way to make this script generic, so we just have to copy-paste the workflow on the next modules.

A solution has been suggested about this, but I haven't checked yet: https://stackoverflow.com/a/58474340

.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          name: ps_checkout
+          path: ../
+      - name: Publish to GitHub Release
+        if: github.event_name == 'release' && github.event.action == 'published'
+        uses: actions/upload-release-asset@v1.0.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.release_info.outputs.upload_url }}
```
Quetzacoalt91: Not sure this is going to work if release-drafter is not run this the same workflow

.github/workflows/build.yml
Quetzacoalt91: Tried on my fork which has a custom name: 
image

.github/workflows/build.yml
PoulainMaxime: I don't know, I just saw on the github action repo that it worked like this.

zalexki: it will because of the event check above, the release is already published

Quetzacoalt91:
Closing to avoid too many notifications

PR #288, opened 2020-02-06
handle icon copy when file does not exist

zalexki:
I feel a little clueless on good practice to do here, do we know if upgrade file can contain use keyword on 1.6 ?
Should i change the scope of setStateIcons method to be able to use it in this upgrade file ?


upgrade/upgrade-1.2.12.php
```diff
@@ -30,5 +32,42 @@
  */
 function upgrade_module_1_2_12($module)
 {
-    return $module->registerHook('displayInvoiceLegalFreeText');
+    foreach (PrestaShop\Module\PrestashopCheckout\OrderStates::ORDER_STATES as $state => $color) {
+        $orderStateId = (int) \Configuration::get($state);
+        setStateIcons($state, $orderStateId, $module->name);
+    }
+
+    return \Db::getInstance()->execute('ALTER TABLE `' . _DB_PREFIX_ . TableManager::TABLE_ORDER_MATRICE . '` CHANGE `id_order_prestashop` `id_order_prestashop` INT(10) UNSIGNED NOT NULL;')
```
Matt75: This come from upgrade file of previous version, it's necessary to run it another time ?

upgrade/upgrade-1.2.12.php
```diff
@@ -30,5 +32,42 @@
  */
 function upgrade_module_1_2_12($module)
 {
-    return $module->registerHook('displayInvoiceLegalFreeText');
+    foreach (PrestaShop\Module\PrestashopCheckout\OrderStates::ORDER_STATES as $state => $color) {
+        $orderStateId = (int) \Configuration::get($state);
```
Matt75: `$orderStateId = (int) \Configuration::getGlobalValue($state);`
OrderState are always global for all shops

zalexki:
@apacios please read description before reviewing most of your questions are what i am asking in the description

upgrade/upgrade-1.2.12.php
zalexki: ye true wasn't sure during rebase, fixed.

upgrade/upgrade-1.2.12.php
zalexki: @Matt75 it's a copy paste of OrderState->setStateIcons, can you please check the description about this.

upgrade/upgrade-1.2.12.php
```diff
@@ -30,5 +30,41 @@
...
+
     return $module->registerHook('displayInvoiceLegalFreeText');
 }
+
+function setStateIcons($state, $orderStateId, $moduleName)
+{
+    $iconExtension = '.gif';
+    $iconToPaste = _PS_ORDER_STATE_IMG_DIR_ . $orderStateId . $iconExtension;
+
+    if (true === file_exists($iconToPaste)) {
```
PoulainMaxime: If file_exists fails it sends a warning exception, wouldn't it be better to do a try catch just in case to avoid an exception?

upgrade/upgrade-1.2.12.php
Matt75: `OrderStates::setStateIcons()` is updated in https://github.com/PrestaShopCorp/ps_checkout/pull/281 to use `\Configuration::getGlobalValue($state);` too

upgrade/upgrade-1.2.12.php
zalexki: php docs says `Upon failure, an E_WARNING is emitted.`

upgrade/upgrade-1.2.12.php
zalexki: ye ok then i should just change method scope and use it there to avoid maintaining it ?

Matt75: Maybe we can extract that to a Command to handle it properly ?
We could then execute it on OrderState class and upgrade file

zalexki: what do you think @v4lux, @apacios ?

upgrade/upgrade-1.2.12.php
```diff
@@ -30,5 +32,42 @@
  */
 function upgrade_module_1_2_12($module)
 {
-    return $module->registerHook('displayInvoiceLegalFreeText');
+    foreach (PrestaShop\Module\PrestashopCheckout\OrderStates::ORDER_STATES as $state => $color) {
+        $orderStateId = (int) \Configuration::get($state);
+        setStateIcons($state, $orderStateId, $module->name);
```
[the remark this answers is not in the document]
zalexki: we can, @Matt75 and me are for it, what do you think ?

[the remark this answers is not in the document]
apacios: @Matt75 @zalexki finaly, as the method `setStateIcons` might be updated in further monthes, we should use it for the upgrade. If the merchant update his module in 1 year we won't have the expected update. 
I think that "duplicating" the code for the upgrade is the best solution. It will never change and we we'll be certain that the result won't change.

upgrade/upgrade-1.2.12.php
v4lux: If it is just a copy/paste it could indeed be better to change the scope of the method setStateIcons().

zalexki: @Quetzacoalt91 can you look at this and above discussion and decide for us ? :D

Quetzacoalt91: I'm in favor in reusing code when possible. In this case, the code is the same for adding or updating the order states.

## Solution 1

We could have the following upgrade:

```php
function upgrade_module_1_2_12($module)
{
    (new PrestaShop\Module\PrestashopCheckout\OrderStates)->updatePaypalStates();

    return $module->registerHook('displayInvoiceLegalFreeText');
}
```

And in OrderStates.php
```php
    /**
     * Update color for each ps_checkout order states
     *
     * @return bool
     */
    public function updatePaypalStates()
    {
        foreach (self::ORDER_STATES as $state => $color) {
            $orderStateId = $this->getPaypalStateId($state, $color);
            $this->setStateIcons($state, $orderStateId);
        }

        return true;
    }
```

## Solution 2

```php
function upgrade_module_1_2_12($module)
{
    (new PrestaShop\Module\PrestashopCheckout\OrderStates)->installPaypalStates();

    return $module->registerHook('displayInvoiceLegalFreeText');
}
```

By calling this method, we run the same code as during the installation process and make sure all order states exist before updating them. If they already exist on the shop, the code simply `continue` to the next order state.

upgrade/upgrade-1.2.12.php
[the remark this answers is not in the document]
Quetzacoalt91: We're partially using the OrderStates class, for instance with its contants. We may go further and use its methods in charge of setting the proper color.

I'm resolving this conversation as we're discussing about it in another one.

upgrade/upgrade-1.2.12.php
```diff
@@ -30,5 +32,42 @@
...
+    }
+
+    return \Db::getInstance()->execute('ALTER TABLE `' . _DB_PREFIX_ . TableManager::TABLE_ORDER_MATRICE . '` CHANGE `id_order_prestashop` `id_order_prestashop` INT(10) UNSIGNED NOT NULL;')
+        && $module->registerHook('displayInvoiceLegalFreeText');
+}
+
+function setStateIcons($state, $orderStateId, $moduleName)
```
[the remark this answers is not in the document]
zalexki: comment nullified from other changes

Quetzacoalt91:
Thank you @zalexki

PR #299, opened 2020-02-24
Add github action in place of travis

PoulainMaxime:


Quetzacoalt91:
We can remove this part from Travis (https://github.com/PrestaShopCorp/ps_checkout/blob/af196b394567dff5e1419a876cc976cc17dd071e/.travis.yml#L47-L82) which will now be in conflict with the new deploy script.

PR #311, opened 2020-02-29
Upgrade PHPStan and extend PrestaShop rules 

Quetzacoalt91:
This PR upgrades PHPStan to v0.12 which includes more rules and provide more feedback, thus the CI broken.

It also updates the way we configure PrestaShop modules specific rules. We now extend a configuration file from `prestashop/php-dev-tools` library, which will define additional rules and fixes ready to be deployed on all modules at the same time.


controllers/front/DispatchWebHook.php
```diff
@@ -36,9 +36,9 @@ class ps_checkoutDispatchWebHookModuleFrontController extends ModuleFrontControl
     public $auth = false;
 
     /**
-     * Id coming from PSL
+     * Id coming from PSL (CheckMe: Is it a UUID?)
```
Quetzacoalt91, line 39: @Matt75 / @v4lux, can you please confirm this property stores a string?

Matt75: Yes, it's a string

Matt75:
@Quetzacoalt91 This PR should be rebased

Quetzacoalt91:
Here you go!

Matt75:
Thanks you @Quetzacoalt91

PR #316, opened 2020-03-05
update order states colors

zalexki:


upgrade/upgrade-1.3.1.php
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Update main function for module Version 1.3.1
+ *
+ * @param Module $module
+ *
+ * @return bool
+ */
+function upgrade_module_1_3_1($module)
+{
+    return (new PrestaShop\Module\PrestashopCheckout\OrderStates())->installPaypalStates();
```
Matt75: No need to add `upgrade/upgrade-1.3.1.php`, action is already triggered by `upgrade/upgrade-1.3.0.php` and we have to make another beta for 1.3.0 so we will add this color changes 👍

PR #320, opened 2020-03-09
Improve queries on module installation for slowest instance

Matt75:
Prevent `MySQL is gone away` on some slow instance cause failure on module installation

`addCheckboxCarrierRestrictionsForModule()` now use one query instead of multiple INSERT queries, `INSERT IGNORE` still used

`addCheckoutPaymentForAllActivatedCountries()` I remove query on each country and use `INSERT IGNORE` to prevent duplicate

### DIFF
--- a/ps_checkout.php
+++ b/ps_checkout.php
@@ -980,30 +980,31 @@ class Ps_checkout extends PaymentModule
      */
     public function addCheckboxCarrierRestrictionsForModule(array $shopsList = [])
     {
-        if (!$shopsList) {
+        if (empty($shopsList)) {
             $shopsList = Shop::getShops(true, null, true);
         }
 
         $carriersList = Carrier::getCarriers((int) Context::getContext()->language->id, false, false, false, null, Carrier::ALL_CARRIERS);
-        $allCarriers = [];
-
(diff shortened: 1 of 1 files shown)


ps_checkout.php
```diff
@@ -980,30 +980,31 @@ public function hookActionFrontControllerSetMedia()
...
-
-        foreach ($carriersList as $carrier) {
-            $allCarriers[] = $carrier['id_reference'];
-        }
+        $allCarriers = array_column($carriersList, 'id_reference');
+        $dataToInsert = [];
 
         foreach ($shopsList as $idShop) {
             foreach ($allCarriers as $idCarrier) {
-                $addModuleInCarrier = Db::getInstance()->execute('INSERT IGNORE
```
apacios: There, I think we used `execute` to be able to use `INSERT IGNORE`.
Are you sure there won't be any issue by removing that ?

Matt75: I still use INSERT IGNORE 👍
See parameters of `Db::getInstance()->insert()` ;)

Quetzacoalt91: That's indeed covered.

PR #323, opened 2020-03-09
Add translations for reporting columns

Matt75:


_dev/src/components/block/reporting.vue
```diff
@@ -105,7 +105,11 @@
               class="label color_field "
               :style="setTransactionBadgeColor(data.item.type)"
             >
-              {{ data.item.type }}
+              {{
+                data.item.type === 'Refund'
+                  ? $t('block.reporting.type.refund')
+                  : $t('block.reporting.type.payment')
+              }}
```
Quetzacoalt91: This data is coming from a Presenter. Its purpose is to make all the calculation so we can reuse the result directly in the templates without additional check. The translations should be called in the class `TransactionPresenter`.

Matt75: Done

PR #324, opened 2020-03-11
Quickfix refund

Matt75:
Quickfix for Refund, it need to be reworked in next major

- Fix notice when call to API fail to retrieve CaptureId
(When order has been placed before logout in ps_checkout and after new onboarding, previous Order are not in the new account, so we cannot retrieve captureId)
- Fix missing save of transaction_id

Problems to be fixed with a refactoring:
- it's not correct to add a new OrderPayment with negative amount, it cause warning on BO Order page
order-warning
- very bad idea to launch a refund on OrderState change, OrderState can change outside BO by others modules. (ERP etc...)

### DIFF
--- a/classes/Refund.php
+++ b/classes/Refund.php
@@ -76,7 +76,19 @@ class Refund
      */
     public function refundPaypalOrder()
     {
-        $response = (new Order(\Context::getContext()->link))->refund($this->getPayload());
+        $payload = $this->getPayload();
+
+        //@todo Quick fix before refactoring
+        if (empty($payload)) {
+            return [
+                'error' => true,
+                'messages' => [
+                    'Unable to retrieve payload or unable to retrieve capture id of this order.',
+                ],
+            ];
+        }
+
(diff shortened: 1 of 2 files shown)


classes/Refund.php
```diff
@@ -92,14 +106,17 @@ public function refundPaypalOrder()
...
             return false;
         }
 
-        $purchaseUnits = current($response['purchase_units']);
-        $capture = current($purchaseUnits['payments']['captures']);
+        //@todo Quick fix before refactoring
+        $purchaseUnits = isset($response['purchase_units']) ? current($response['purchase_units']) : null;
+        //@todo Quick fix before refactoring
+        $capture = isset($purchaseUnits['payments']['captures']) ? current($purchaseUnits['payments']['captures']) : null;
         $captureId = $capture['id'];
```
Quetzacoalt91: This is going to throw errors if `$capture` is null.

Matt75: Done

PR #326, opened 2020-03-12
Continuous delivery

MaxencePerrinPrestashop:
- [x] Continuous delivery to integration shop on push to dev
- [x] Continuous delivery to production shop on push to prod

### DIFF
--- a/.github/workflows/checkout-cd-inte.yml
+++ b/.github/workflows/checkout-cd-inte.yml
@@ -1,0 +1,85 @@
+name: Module checkout CD
+
+on:
+  push:
+    branches:
+      - dev
+
+env:
+  MODULE_NAME: ps_checkout
+  GCLOUD_TOKEN_PATH: ./token.json
+
+jobs:
+  checkout_cd:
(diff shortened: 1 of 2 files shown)


.github/workflows/checkout-cd-inte.yml
```diff
@@ -0,0 +1,86 @@
...
+env:
+  MODULE_NAME: ps_checkout
+  GCLOUD_TOKEN_PATH: ./token.json
+
+jobs:
+  checkout_cd:
+    name: Module Checkout continuous deployment
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    if: contains(github.event.pull_request.labels.*.name, 'quality insurance needed')
```
Quetzacoalt91: Is this check going to work? If you push to `dev`, there shouldn't be any more `pull_request` array to rely on?

MaxencePerrinPrestashop: Thx for the review ! I will change the trigger to be pull_request.

PR #329, opened 2020-03-12
Quick fix transaction id

Matt75:
Fix mixed content of transaction_id was "PayPal Order Id instead" of "PayPal Transaction Id"
Add some logs to find root cause of some fails

Many API Call can fail without clear informations...

This part need to be refactored

### DIFF
--- a/classes/ValidateOrder.php
+++ b/classes/ValidateOrder.php
@@ -67,22 +67,40 @@ class ValidateOrder
      */
     public function validateOrder($payload)
     {
+        // API call here
         $paypalOrder = new PaypalOrder($this->paypalOrderId);
         $order = $paypalOrder->getOrder();
+
+        if (empty($order)) {
+            // @todo quickfix : Call API return nothing or fail
+            $message = sprintf('Unable to retrieve Paypal Order for %s', $this->paypalOrderId);
+            \PrestaShopLogger::addLog($message, 1, null, null, null, true);
+            throw new \PrestaShopException($message);
+        }
+
         $apiOrder = new Order(\Context::getContext()->link);
 
         switch ($paypalOrder->getOrderIntent()) {
             case self::INTENT_CAPTURE:
+                // API call here
                 $response = $apiOrder->capture($order['id'], $this->merchantId);
                 break;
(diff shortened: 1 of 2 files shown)


controllers/front/ValidateOrder.php
```diff
@@ -73,12 +74,12 @@ public function postProcess()
             'cartId' => (int) $cart->id,
             'amount' => $total,
             'paymentMethod' => $paymentMethod,
-            'extraVars' => ['transaction_id' => $paypalOrderId],
```
Quetzacoalt91: Isn't it interesting to have the transaction ID in the payment details? Even if we display it later in another table, this information is really important for the merchant

Matt75: It's not module->validateOrder() here ;)

Quetzacoalt91: Yeah, this information is not used anymore regarding the changes in `classes/ValidateOrder.php`

classes/ValidateOrder.php
```diff
@@ -67,22 +67,40 @@ public function __construct($paypalOrderId, $merchantId)
...
             case self::INTENT_AUTHORIZE:
+                // API call here
                 $response = $apiOrder->authorize($order['id'], $this->merchantId);
                 break;
             default:
-                throw new \Exception(sprintf('Unknown Intent type %s', $paypalOrder->getOrderIntent()));
+                // @todo quickfix
+                $message = sprintf('Unknown Intent type %s, Paypal Order %s', $paypalOrder->getOrderIntent(), $this->paypalOrderId);
+                \PrestaShopLogger::addLog($message, 1, null, null, null, true);
+                throw new \PrestaShopException($message);
```
Quetzacoalt91: What is the reason for using `PrestaShopException` instead of `Exception`?

In a later version, we could implement our own `PsCheckoutException` who would handle the logging by itself.

Matt75: Due to this https://github.com/PrestaShopCorp/ps_checkout/blob/master/controllers/front/DispatchWebHook.php#L113

Quetzacoalt91: All right!

PR #334, opened 2020-03-17
Avoid JS error Uncaught (in promise)

Matt75:
Avoid a JS error Uncaught (in promise)
error-uncaught-promise
Cause trouble on some browser and stop navigation to finish Onboarding for example

See https://github.com/vuejs/vue-router/issues/2873


NeOMakinG:
You should comment it btw, so someone doesn't think wrong about it later

PR #342, opened 2020-04-01
Fix webhook sent json encoded data

Matt75:
Sometime we have a `stdClass` object instead of an `array` because `resource` is still json encoded in webhook payload

Cause an error fatal and HTTP 500 response

### DIFF
--- a/classes/webHookDispatcher/OrderDispatcher.php
+++ b/classes/webHookDispatcher/OrderDispatcher.php
@@ -61,13 +61,13 @@ class OrderDispatcher implements Dispatcher
         }
 
         if ($payload['eventType'] === self::PS_CHECKOUT_PAYMENT_REFUNED
-        || $payload['eventType'] === self::PS_CHECKOUT_PAYMENT_REVERSED) {
+            || $payload['eventType'] === self::PS_CHECKOUT_PAYMENT_REVERSED) {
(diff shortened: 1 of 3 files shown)


classes/webHookDispatcher/OrderDispatcher.php
```diff
@@ -55,20 +55,22 @@ public function __construct()
     public function dispatchEventType($payload)
     {
         $psOrderId = $this->getPrestashopOrderId($payload['orderId']);
+        // Some webhook (queued) sent json encoded data
+        $resource = is_array($payload['resource']) ? $payload['resource'] : \Tools::jsonDecode($payload['resource'], true);
```
Quetzacoalt91, line 59: Note you don't have to use anymore `Tools::jsonEncode` and you're free to use the native method `json_encode`

Matt75: With PrestaShop 1.6 too ?

Quetzacoalt91: On all PrestaShop versions, this has been removed from the validation rules.

Matt75: I will do it in another PR, there are many `Tools::jsonDecode()`elsewhere

Quetzacoalt91:
Thank you @Matt75

PR #349, opened 2020-04-16
Introduce Service Container with Dependency Injection

Matt75:
To start refactoring we need a Service Container to make Dependency Injection to facilitate introducing of new Services and automated Tests.

Since PrestaShop 1.7.3 there are in fact 3 containers :
- legacy front container available on front office of the shop.
- legacy admin container available on administration of the shop.
- Symfony container available on administration of the shop **but only for migrated controllers**.

So in our case we are always in legacy so we don’t have access to Symfony container so have to instanciate ourself services we need.

Due to support of PrestaShop 1.6.1 and first versions of PrestaShop 1.7, we have to build our own and start PrestaShop 1.7.3 we can use one provided by Core.

Some services will be common to front and back, like Cache or Logger Services for example and some are limited to administration like refund and reporting services.

I added an unit test for new CacheDirectory Service, responsible for returning cache directory path used to store Service Container cache before PrestaShop 1.7.3

New Logger Service will be introduced in another Pull Request


classes/Configuration/Configuration.php
```diff
@@ -0,0 +1,125 @@
...
+    {
+        return 15;
+    }
+
+    /**
+     * @return string
+     */
+    public function getPrestaShopPath()
+    {
+        return (string) constant('_PS_ROOT_DIR_');
```
Quetzacoalt91, line 51: I guess you use the method `constant` to prevent PHP warnings, but when we look at php.net, it seems the methods triggers a warning anyway : https://www.php.net/manual/en/function.constant.php

Matt75: Fixed 👍

classes/Logger/LoggerPathProvider.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public function __construct(Configuration $configuration)
+    {
+        $this->configuration = $configuration;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPath()
```
Quetzacoalt91, line 46: As the role of this class is to only provide the path of the logger, which is already specified in the class name, you can simply call your method get().

However, if you think this class could make more than providing the path, keep the method `getPath` but update the class name with something more generic.

classes/DependencyInjection/ContainerProvider.php
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * @param string $containerName
+     *
+     * @return ContainerInterface
+     */
+    public function getContainer($containerName)
+    {
+        $containerClassName = ucfirst($this->configuration->getModuleName())
+            . (false === empty($containerName) ? ucfirst($containerName) : '')
```
Quetzacoalt91, line 62: If we always expect a string, I guess `empty($containerName)` is already an empty string?

classes/DependencyInjection/ContainerProvider.php
```diff
@@ -0,0 +1,90 @@
...
+            . 'Container'
+        ;
+        $containerFilePath = $this->cachePathProvider->getPath() . '/' . $containerClassName . '.php';
+        $containerConfigCache = new ConfigCache($containerFilePath, $this->configuration->isDebug());
+
+        if ($containerConfigCache->isFresh()) {
+            require_once $containerFilePath;
+
+            return new $containerClassName();
```
Quetzacoalt91, line 71: I see that you're using the class `Configuration`, and define it in the services afterwards. I wonder if this class won't be instantiated two times.
This seems acceptable today as the class do not execute heavy tasks in its constructor, but we may encounter issues later with further changes.

We could avoid instantiation of the same class by affecting the instance we already have in hand:

```php
$containerBuilder->set('configuration', $this->configuration);
```

Matt75: Applicable only if we build our own container, if we are on PrestaShop 1.7.3+ Core container will not register them...

Quetzacoalt91: Of course, keep its definition in your services.yml files, here is a a shortcut to reuse an existing instance instead of creating another one.

classes/Configuration/Configuration.php
```diff
@@ -0,0 +1,183 @@
...
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\Module\PrestashopCheckout\Configuration;
+
+use Module;
+use Psr\Log\LogLevel;
+
+class Configuration
```
Quetzacoalt91, line 26: If you're still to rename this class so there's no misunderstanding with the core class, there is something like `RuntimeEnvironment` which could be a fit

classes/Cache/CachePathProvider.php
```diff
@@ -0,0 +1,55 @@
...
+ * @copyright 2007-2020 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ * International Registered Trademark & Property of PrestaShop SA
+ */
+
+namespace PrestaShop\Module\PrestashopCheckout\Cache;
+
+use PrestaShop\Module\PrestashopCheckout\Configuration\Configuration;
+
+class CachePathProvider implements CachePathProviderInterface
```
zalexki, line 25: looks like you could use a ProviderInterface used accross all providers don't you think ?

Matt75: True, interfaces needs some polish.

Factory\CheckoutLogger will be removed in favor of new LoggerFactory in this PR.

classes/Configuration/Configuration.php
Matt75: I need data from .env, constants, PrestaShop Configuration, PrestaShop Context, Module...
I will try somethings with [symfony/config](https://symfony.com/doc/current/components/config.html) and [symfony/dotenv](https://symfony.com/doc/current/components/dotenv.html)

Matt75:
⚠️ To be merged on a new branch v2.0.0 cause branch dev is for v1 maintenance for now

classes/Cache/CacheDirectory.php
```diff
@@ -0,0 +1,98 @@
...
+     */
+    private $psVersion;
+
+    /**
+     * @var string PrestaShop path
+     */
+    private $psPath;
+
+    /**
+     * @var string PrestaShop Debug Mode
```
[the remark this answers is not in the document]
Matt75: Nice catch

PR #353, opened 2020-04-23
Improve PayPal Sdk link building

Matt75:
Quickwin to be refactored in v2.0.0


controllers/front/PaymentCard16.php
```diff
@@ -54,16 +54,19 @@ public function initContent()
...
-        if (false === $paypalAccountRepository->cardPaymentMethodIsValid()) {
+        if (false === $paypalAccountRepository->cardHostedFieldsIsAvailable()) {
             $this->redirectToHomePage();
         }
 
         $paypalOrder = new CreatePaypalOrderHandler($this->context);
         $paypalOrder = $paypalOrder->handle();
 
         $language = (new LanguageAdapter())->getLanguage($this->context->language->id);
+        $paypalSdkLink = new PrestaShop\Module\PrestashopCheckout\Builder\PayPalSdkLink\PayPalSdkLinkBuilder();
```
v4lux, line 65: We are not in the main class of the module. It will be better to import the class using the use statement.

controllers/front/PaymentPaypal16.php
```diff
@@ -62,8 +62,11 @@ public function initContent()
         $paypalOrder = $paypalOrder->handle();
 
         $language = (new LanguageAdapter())->getLanguage($this->context->language->id);
+        $paypalSdkLink = new PrestaShop\Module\PrestashopCheckout\Builder\PayPalSdkLink\PayPalSdkLinkBuilder();
```
v4lux, line 65: Same here.

classes/Builder/PayPalSdkLink/PayPalSdkLinkBuilder.php
```diff
@@ -0,0 +1,205 @@
...
+        $fundingSourcesDisabled = $this->getFundingSourcesDisabled();
+
+        if ($this->isSmartButtonsEnabled() && false === empty($fundingSourcesDisabled)) {
+            $params['disable-funding'] = implode(',', $this->getFundingSourcesDisabled());
+        }
+
+        return self::BASE_LINK . '?' . urldecode(http_build_query($params));
+    }
+
+    public function displayExpressCheckout()
```
Quetzacoalt91, line 104: To aviod confusion between a setter and a method display content on a page, what do you think about having `enable` as a verb?

```suggestion
    public function enableDisplayExpressCheckout()
```

classes/Repository/PaypalAccountRepository.php
```diff
@@ -184,13 +184,167 @@ public function getPaypalPaymentStatus()
...
+    public function isSofortEnabled()
+    {
+        return (bool) \Configuration::get(
+            PaypalAccount::PS_CHECKOUT_FUNDING_SOURCE_SOFORT_ENABLED,
+            null,
+            null,
+            (int) \Context::getContext()->shop->id,
+            true
+        );
+    }
```
Quetzacoalt91, line 349: The idea of having one method and constant for each payment method is good, so we can for instance enjoy autocompletion with the IDE, but there is duplicated we can avoid.

We could have a generic method:

```php
    /**
     * Generic method checking a payment method is enabled
     *
     * @return bool
     */
    public function isPaymentMethodEnabled($paymentMethod)
    {
        return (bool) \Configuration::get(
            $paymentMethod,
            null,
            null,
            (int) \Context::getContext()->shop->id,
            true
        );
    }
```

Then you can reuse it on all the other methods. For instance:

```php
   /**
     * @return bool
     */
    public function isSofortEnabled()
    {
        return $this->isPaymentMethodEnabled(PaypalAccount::PS_CHECKOUT_FUNDING_SOURCE_SOFORT_ENABLED);
    }
```

This allows you to fix only one method in the future if the way of checking a payment method evolves.

PR #356, opened 2020-04-27
Hide shop context panel with multishop in configuration page

Matt75:
ps_checkout_configuration

### DIFF
--- a/views/templates/admin/configuration.tpl
+++ b/views/templates/admin/configuration.tpl
@@ -25,7 +25,7 @@
 
 <style>
   /** Hide native multistore module activation panel */
-  #content div.bootstrap.panel {
+  #content.nobootstrap div.bootstrap.panel {
     display: none;
   }
 </style>
--- a/views/templates/admin/configuration.tpl
+++ b/views/templates/admin/configuration.tpl
@@ -24,7 +24,7 @@
 <script src="{$pathApp|escape:'htmlall':'UTF-8'}"></script>
 
 <style>
-  /** Hide native multistore module activation panel */
+  /** Hide native multistore module activation panel, because of visual regressions on non-bootstrap content */
   #content.nobootstrap div.bootstrap.panel {
     display: none;
   }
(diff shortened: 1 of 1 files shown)


views/templates/admin/configuration.tpl
```diff
@@ -22,3 +22,10 @@
 <div id="app"></div>
 
 <script src="{$pathApp|escape:'htmlall':'UTF-8'}"></script>
+
+<style>
+  /** Hide native multistore module activation panel */
+  #content div.bootstrap.panel {
```
Quetzacoalt91, line 28: This seems to be a generic selector, this could have some side effects if other modules are displayed on this page?

Matt75: It's the PrestaShop Checkout configuration page so no other modules...
PrestaShop Checkout component are in #app and not affected by this.

Quetzacoalt91: This is what you consider by default on a shop, but PrestaShop provides a hook which could allow modules to add content at the end of the page (you have heard of `displayAdminEndContent` with ps_mbo :slightly_smiling_face:).

If someone decided to add content and reusing the generic classes of Bootstrap, there is a risk that you alter other modules content. This solution is not as clean a yours, but it prevents side effects:

```suggestion
  $('input[name="activateModule"]', '#content').parent('.panel')
```

Matt75: I did that due to PrestaShop Core bug not handle correctly nobootstrap class.
We don't want any other modules to add content on this page with unattended display.
If you add content reusing the generic classes of Bootstrap it will not work well due to nobootstrap, so it's attended to hide them.

views/templates/admin/configuration.tpl
```diff
@@ -22,3 +22,10 @@
 <div id="app"></div>
 
 <script src="{$pathApp|escape:'htmlall':'UTF-8'}"></script>
+
+<style>
+  /** Hide native multistore module activation panel */
```
Quetzacoalt91, line 27: Please provide an explanation on the reason you hide this. Something like:

```suggestion
  /** Hide native multistore module activation panel because visual regressions on non-bootstrap content */
```

PR #359, opened 2020-04-28
Fix unsaved PayPal Order will cause fail of webhook

Matt75:
Always save PayPal Order after a PrestaShop Order is created to avoid a webhook fail due to not found order.
`Error: order [PAYPAL ORDER] does not exist`

Due to PrestaShop Core split sometime a Cart in multiple Orders : https://github.com/PrestaShop/docs/issues/508


Quetzacoalt91:
Thank you @Matt75

PR #362, opened 2020-05-05
Improve Order View : asynchronous request, support of PrestaShop 1.7.7

Matt75:
- Fix performance issue on Order View due to synchronous request (API Call to fetch Order) => use an asynchronous request with ajax
- Fix translation issue : function `l` of smarty not support `sprintf` in PrestaShop 1.6
- Add support of new Order page on PrestaShop 1.7.7 => You should have latest nightly with hook `DisplayAdminOrderMainBottom`


views/templates/hook/displayAdminOrderLeft.tpl
```diff
@@ -22,153 +22,31 @@
...
+      orderPayPalRefundButton: '#ps_checkout button.refund',
+      orderPayPalModalContainerPrefix: '#ps-checkout-refund-',
+      orderPayPalModalContainer: '.ps-checkout-refund',
+      orderPayPalModalNotificationsContainer: '.modal-notifications',
+      orderPayPalModalContentContainer: '.modal-content-container',
+      orderPayPalModalLoaderContainer: '.modal-loader',
+      orderPayPalModalRefundForm: '.ps-checkout-refund-form',
     });
-  });
+  }
```
Quetzacoalt91, line 51: To avoid duplicated code between this file and views/templates/hook/displayAdminOrderMainBottom.tpl, can you move this part in another file which can be included? https://www.smarty.net/docsv2/en/language.function.include.tpl

upgrade/upgrade-1.4.0.php
```diff
@@ -143,6 +143,8 @@ function upgrade_module_1_4_0($module)
 
     return $result
         && $module->registerHook('displayAdminOrderLeft')
+        && $module->registerHook('displayAdminOrderMainBottom')
```
Quetzacoalt91, line 146: This hook is in your list for PS 1.7, don't you want to check the current version?

Matt75: Not needed, it will be added be INSERT IGNORE on PrestaShop upgrade so no conflict.
If a merchant upgrade from an older version, no need to register hook manualy

views/templates/hook/displayAdminOrderLeft.tpl
Matt75: How manage `legacy` argument not the same ?

upgrade/upgrade-1.4.0.php
Quetzacoalt91: That's a fair point, but in that case why don't we register it for new merchants too? You should have the same behavior for everyone or you will encounter issues depending on the life of the module on the shop.

Matt75: Need to check all hooks in upgrade script of PrestaShop because some are added without INSERT IGNORE and cause crash of PrestaShop upgrade (mainly fist versions of 1.7)

views/templates/hook/displayAdminOrderLeft.tpl
Quetzacoalt91: you may send it as another parameter

PR #364, opened 2020-05-05
Add support of outofstock order state

Matt75:
PrestaShop use OrderState `On backorder (not paid)` and `On backorder (paid)` if a product is outofstock.

See https://github.com/PrestaShop/PrestaShop/blob/1.7.6.x/classes/PaymentModule.php#L559

### How to test
- Create a Cart with a product outofstock but allowed to be ordered.
- Go to Proceed to checkout
- Choose to paid with PrestaShop Checkout
- Paid
- Go to BO > Orders
- See OrderState `On backorder (paid)` instead of `Payment accepted`

### DIFF
--- a/classes/webHookDispatcher/OrderDispatcher.php
+++ b/classes/webHookDispatcher/OrderDispatcher.php
@@ -30,24 +30,6 @@ class OrderDispatcher implements Dispatcher
     const PS_CHECKOUT_PAYMENT_DENIED = 'PaymentCaptureDenied';
 
     /**
-     * @var array
-     */
-    private $matriceEventAndOrderState;
-
-    /**
-     * OrderDispatcher constructor.
-     */
-    public function __construct()
-    {
-        $this->matriceEventAndOrderState = [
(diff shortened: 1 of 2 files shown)


classes/webHookDispatcher/OrderDispatcher.php
```diff
@@ -191,4 +173,56 @@ private function dispatchPaymentStatus($eventType, $resource, $orderId)
...
+     * @return int OrderState paid identifier
+     */
+    private function getPaidStatusId($currentOrderStateId)
+    {
+        if ($currentOrderStateId === (int) \Configuration::getGlobalValue('PS_OS_OUTOFSTOCK_UNPAID')) {
+            return (int) \Configuration::getGlobalValue('PS_OS_OUTOFSTOCK_PAID');
+        }
+
+        return (int) \Configuration::getGlobalValue('PS_OS_PAYMENT');
+    }
```
Quetzacoalt91, line 213: This method is too close to `getPaidStatusId` in classes/ValidateOrder.php and must be merged in the next version with it to avoid duplicated code.

Matt75: Will be a part of a new Service in 2.0.0 :)

PR #379, opened 2020-05-12
Fix PSX form response

Matt75:
Fix regression from https://github.com/PrestaShopCorp/ps_checkout/pull/378

### DIFF
--- a/controllers/admin/AdminAjaxPrestashopCheckoutController.php
+++ b/controllers/admin/AdminAjaxPrestashopCheckoutController.php
@@ -219,7 +219,7 @@ class AdminAjaxPrestashopCheckoutController extends ModuleAdminController
 
         // Save form in database
         if (false === $this->savePsxForm($psxForm)) {
-            $this->ajaxDie(json_encode(false));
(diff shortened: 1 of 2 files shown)


controllers/admin/AdminAjaxPrestashopCheckoutController.php
```diff
@@ -219,7 +219,7 @@ public function ajaxProcessPsxSendData()
 
         // Save form in database
         if (false === $this->savePsxForm($psxForm)) {
-            $this->ajaxDie(json_encode(false));
+            $this->ajaxDie(json_encode(['Cannot save in database.']));
```
apacios, line 222: Shouldn't it be translatable ?

Matt75: Nope, this case should never happens and this whole shit will be gone in v2.0.0

Matt75: It save with Configuration::updateValue never return false, can return 0, 1 or true ^^

PR #381, opened 2020-05-18
Rename function name to avoid conflicts

Quetzacoalt91:
Following the feedback on the forum: https://www.prestashop.com/forums/topic/1023519-fatal-error-module-file/


Matt75:
Thanks you @Quetzacoalt91

PR #382, opened 2020-05-19
Fix Cart Presenter - Customer not found in some case

Matt75:
If guest checkout is enabled in BO > Shop Parameters > Order settings, sometimes we need to find guest created in `Customer` with `id_customer` of `Cart` instead of `Context` to retrieve email.
```
Invalid request payload input. child "Payer" fails because
[child "email_address" fails because ["email_address" must be a string]].
{"value":null,"key":"email_address","label":"email_address"}
````

### DIFF
--- a/classes/Presenter/Cart/CartPresenter.php
+++ b/classes/Presenter/Cart/CartPresenter.php
@@ -21,8 +21,6 @@
 namespace PrestaShop\Module\PrestashopCheckout\Presenter\Cart;
 
 use PrestaShop\Module\PrestashopCheckout\Presenter\PresenterInterface;
-use PrestaShop\Module\PrestashopCheckout\ShopContext;
-use PrestaShop\PrestaShop\Adapter\Cart\CartPresenter as PsCartPresenter;
 
 /**
  * Present the cart waiting by the create order paypal builder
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: classes/Presenter/Cart/CartPresenter.php [hunk 1]
```diff
@@ -21,138 +21,56 @@
...
+        $context = \Context::getContext();
+        $productList = $context->cart->getProducts();
 
-        $cart = (array) $this->getCart();
+        $cart = (array) $context->cart;
 
-        if ((new ShopContext())->isShop17()) {
-            $cart = new PsCartPresenter();
-            $cart = $cart->present($this->getCart());
+        if (class_exists('\PrestaShop\PrestaShop\Adapter\Cart\CartPresenter')) {
```
Quetzacoalt91, line 44: I'm curious, why have you chosen this test ?

Matt75: Avoid to create another instance of ShopContext not required because we could only check if class is declared
In v2.0.0 we don't use CartPresenter anymore

Matt75: And the `use` at beginning of files of a class not exist on 1.6 can be a problem, that's why I removed it in favor of full namespace

PR #385, opened 2020-05-19
Compability module: onepagecheckoutps - PresTeamShop

presteamshop:
Hello prestashop team,

They are two small changes that do not affect the operation of your module, but make it compatible with our onepagecheckoutps module. We have many clients with our module that do not use ps_checkout because both modules are incompatible, for this reason we need these changes to be taken into account, just as we have also made several changes to our module to make it compatible.

our mail: info@presteamshop.com

Best regards.


Matt75:
@presteamshop Thanks you for your contribution 👍

I tested your Pull Request with onepagecheckoutps v4.0.9 on PrestaShop 1.7.6.5 and PayPal SDK is not loaded so PrestaShop Checkout is not working :

onepagecheckoutps

We are working on v2.0.0 currently and we will refactoring JS part of the module and Tpl files to improve PrestaShop Checkout compatibility with OnePageCheckout modules. 👍

presteamshop:
Thank you for your prompt response!

Compatibility with ps_checkout will now be released in version 4.0.10 :)

If you send me an email, I can send the .zip with this compatibility.

PS: I would like to contact a previous version of 2.0.0 to leave both modules working before launch.

Best regards.

Matt75:
I will send you an email, thanks 👍

Matt75:
@presteamshop Hi, I tested your module onepagecheckoutps
 v4.0.9 that you sent me by email with [ps_checkout v1.4.1](https://github.com/PrestaShopCorp/ps_checkout/milestone/4)

You are not using PrestaShop standard HTMLElement classes for PaymentOption so our module cannot work properly.
onepagecheckoutps-paymentoption

You should use same identifier than PrestaShop Core to ensure compatibility with all Payment Module.
core-paymentoption

presteamshop:
@Matt75 thanks for you replay.

It is correct that we do not use the same structure of the native prestashop, but we do keep certain identifiers and classes. We have added the identifier "{$ option.id}-container" which we did not have, thanks to your observation.

Can you export us version 1.4.1 for us to try? It is that with the implementation of the "docker" we have not been able to work the branch "dev" ourselves, since from windows we did not find a solution to compile (it is interesting if you know how to do it and can guide us).

We look forward to continue with this compatibility.

Thank you.

Best regards.

Matt75:
Hi @presteamshop

Release builder is here : https://github.com/PrestaShopCorp/ps_checkout/blob/master/.github/workflows/build-release.yml

I will sent you by email a zip for the current release draft.

By the way, I think compatibility with v1.4.1 will be hard due to https://github.com/PrestaShopCorp/ps_checkout/blob/7dbf9729f65f8c3197ec0131244da86b1b8e1ad8/ps_checkout.php#L446

In v2.0.0 we will remove that and generate PayPal order with asynchronous Ajax request but we cannot change that in v1 due to important changes required.

PR #389, opened 2020-05-22
[WIP] New webhook system from accounts

xGouley:


ps_checkout.php
```diff
@@ -137,11 +137,14 @@ public function __construct()
      */
     public function install()
     {
+        $webhookHelper = new PrestaShop\AccountsAuth\Wrapper\Webhook();
```
xGouley, line 140: La lib est-elle bien mise en dep, et publiée ? Prérequis !

ps_checkout.php
```diff
@@ -360,9 +363,15 @@ public function hookDisplayFooterProduct($params)
 
     public function getContent()
     {
+        $webhookHelper = new PrestaShop\AccountsAuth\Wrapper\Webhook();
+        $webhookHelper->ensureWebhookRegistered($this);
```
xGouley, line 367: Sert à enregistrer le Hook meme après l'install, pour faire fonctionner PS Checkout legacy correctement.

ps_checkout.php
```diff
@@ -1044,4 +1054,16 @@ public function hookDisplayAdminOrderMainBottom(array $params)
 
         return $this->display(__FILE__, '/views/templates/hook/displayAdminOrderMainBottom.tpl');
     }
+
+    public function hookReceiveWebhook_ps_checkout(array $params)
```
xGouley, line 1058: Nom de la fonction très important et totalement régit par "hookReceiveWebhook_${module->name}"

ps_checkout.php
david-piatek: yep https://github.com/PrestaShopCorp/prestashop_accounts_auth/tree/feature/790-extract-library-from-ps_accounts

ps_checkout.php
david-piatek: du coup idéalement dans le [README de ps_accounts](https://github.com/PrestaShopCorp/ps_accounts/edit/master/Readme.md)

ps_checkout.php
Quetzacoalt91: As we don't know what is the time needed to run it each time we load the page, it could be enough to move this code in a upgrade file.

ps_checkout.php
Quetzacoalt91: There are other way to specify what module should be targeted by a specific module. What do you think about restricting the hook names we could register (i.e `hookActionReceiveWebhook`) ?

### Solution 1: Restrict the module on the caller side

When calling Hook::exec, there is a parameter which allows you to filter what module will be called.

*classes/Hook.php:*
```php
    public static function exec(
        $hook_name,
        $hook_args = [],
        $id_module = null, // <-- The param I'm talking about
        $array_return = false,
        $check_exceptions = true,
        $use_push = false,
        $id_shop = null,
        $chain = false
```

## Solution 2:

Allow the hook to be dispatched everywhere, but send the primary target as a parameter.

When triggering the hook, we could have the main module in the args:

```php
$hookArgs = [
    'service' => 'ps_checkout',
    // ...
];

Hook::exec('actionReceiveWebhook', $hookArgs);
```

Then, on the ps_checkout side:

```php
public function hookActionReceiveWebhook(array $params)
{
    if ($params['service'] !== $this->name) {
        return;
    }
}
```

ps_checkout.php
Quetzacoalt91: To make sure we may test it, the lib should be added into the composer.json

Matt75:
@xGouley Please target branch v2.0.0

xGouley:
Pour le moment, exemple suspendu, on ne sait pas encore si on va utiliser ce mécanisme.

PR #390, opened 2020-05-22
Improve exception, webhook http code, error handling on validateOrder

Matt75:
Improve error handling :
1. Handle processing payment error smoothly
2. Notify Customer Service on processing payment error cannot be resolved by Customer
3. Avoid Fatal Error on controllers DispatchWebhook and ValidateOrder

This should be released on 1.5.0 cause it will required tests from QA and new translations.

If processing payment error can be resolved by Customer :
customer-error

If processing payment error cannot be resolved by Customer because it required action from Merchant, we display an error page with contact form or a link to Contact Form (depends if contactform widget is available) : 
validate-order-error

We create a CustomerService Thread on this case
customer-thread


Darmona:
Hello,
3 questions :
1 - Can we select the radio button that was used just before? (error when paying by card re-display hosted card fields to try another card for instance)
2 - When the transaction has succeeded but action from the merchant needed, the message is too much negative and make customer anxious, can we change it? (I can provide wording)
3 - Do ALL cases of error are included in the 2 first cases:
- either customer needs to pay again => redirect to the payment step,
- either payment has been through successfully but orders not created then merchant action needed.
Are we missing some corner cases? 
4 - What customer name is sending the customer service message? (here john doe) If an order is created, can we display order reference in message? 
5 - If I have the Merchant Action Needed screen and I go to customer service to send a message, my message will be displayed in the same thread as the automatic message sent by the module ?

Matt75:
@Darmona 
1. We will be able to do that in v2.0.0 but not in v1.5.0 due to redirection
2. Yes, I will change the message
3. I checked all cases on PayPal docs, but there can be undocumented case.
4. The customer name sent to Customer Service in the one used on Order process. We cannot add an Order reference because no order is created at this step, only a Cart.
5. No because the automatic message is sent with flag private due to error details may contain debug information with sensible data.

Matt75:
@Darmona Wordings updated, added ContactForm after error message only on 1.7 and if contactform module is enabled.

classes/webHookDispatcher/OrderDispatcher.php
```diff
@@ -95,22 +98,22 @@ private function getPrestashopOrderId($orderId)
...
-        $initiateBy = 'Merchant';
+        (new WebHookValidation())->validateRefundResourceValues($resource);
 
-        if ($eventType === self::PS_CHECKOUT_PAYMENT_REVERSED) {
-            $initiateBy = 'Paypal';
-        }
+        return true;
 
-        return (new WebHookOrder($initiateBy, $resource, $orderId))->updateOrder();
+//        $initiateBy = 'Merchant';
```
zalexki, line 110: remove commented code

ps_checkout.php
```diff
@@ -1039,20 +1039,77 @@ public function hookDisplayAdminOrderMainBottom(array $params)
      */
     public function hookDisplayPaymentTop()
```
apacios, line 1040: I think it would be best to put hooks in classes to make the main class easier to read and understand. Over 1000 lines it's starting to do.

PR #400, opened 2020-05-29
Fix Uncaught SyntaxError: Unexpected token p in JSON

Matt75:
Uncaught SyntaxError: Unexpected token p in JSON at position 523
at JSON.parse (<anonymous>)
at initPsCheckout (v_157_13126d94a58fdcb11d400f3873333ff2.js:733)
at v_157_13126d94a58fdcb11d400f3873333ff2.js:732

### DIFF
--- a/views/templates/front/expressCheckout.tpl
+++ b/views/templates/front/expressCheckout.tpl
@@ -46,43 +46,40 @@
   }
   #paypal-button-container {
     max-width:300px;
+  }
+  .page-cart #paypal-button-container {
     margin: auto;
   }
 </style>
 
 <script>
-/**
- * Load paypal script
- */
-function loadPaypalScript() {
-  if (typeof paypalSdkPsCheckout !== 'undefined') {
-    return;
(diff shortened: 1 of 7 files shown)


views/templates/front/expressCheckout.tpl
```diff
@@ -46,43 +46,40 @@
...
-/**
- * Load paypal script
- */
-function loadPaypalScript() {
-  if (typeof paypalSdkPsCheckout !== 'undefined') {
-    return;
-  }
-
-  let psCheckoutScript = document.getElementById('psCheckoutPaypalSdk');
+  const checkoutLink = "{$checkoutLink|escape:'javascript':'UTF-8' nofilter}";
```
Quetzacoalt91, line 56: This should be `|escape:'javascript':'UTF-8'` or `nofilter`, but both do not make sense.

Matt75: It's required because on 1.7 PrestaShop apply escaping + apply escape

Matt75: Try with and without ;)

PR #421, opened 2020-06-16
Fix payment method label

Matt75:
Fix link used for Buttons


ondo27:
Yes, the solution is proved. Thank you guys

Matt75:
@ondo27 Thanks for your feedback 👍
@Quetzacoalt91 Thanks for review 💪

PR #429, opened 2020-06-29
Adds logs on ValidateOrder exception

Matt75:
Add a notice log for exception can be resolved by customer himself ($notifyCustomerService = false)
Add an error log for exception cannot be resolved by customer himself (send message to CustomerService)

### DIFF
--- a/controllers/front/ValidateOrder.php
+++ b/controllers/front/ValidateOrder.php
@@ -220,6 +220,11 @@ class ps_checkoutValidateOrderModuleFrontController extends ModuleFrontControlle
         $exceptionMessageForCustomer = $this->module->l('Error processing payment', 'translations');
         $exceptionCode = $exception->getCode();
         $notifyCustomerService = true;
+        $paypalOrder = Tools::getValue('orderId');
+
+        if (false === Validate::isGenericName($paypalOrder)) {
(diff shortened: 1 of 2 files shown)


controllers/front/ValidateOrder.php
```diff
@@ -241,7 +246,7 @@ private function handleException(Exception $exception)
                 $this->redirectToCheckout(['step' => 'payment', 'paymentError' => $exception->getCode()]);
                 break;
             case PayPalException::ORDER_ALREADY_CAPTURED:
-                $this->module->currentOrder = (new \OrderMatrice())->getOrderPrestashopFromPaypal(Tools::getValue('orderId'));
+                $this->module->currentOrder = (new \OrderMatrice())->getOrderPrestashopFromPaypal($paypalOrder);
```
Quetzacoalt91, line 249: I haven't seen the specific case 'invalid' to be handled here or in `getOrderPrestashopFromPaypal`. This means we might find a PrestaShop Order and redirect the customer to a wrong OrderConfirmation, don't you think ?

Matt75: Cannot be happens, PsCheckoutException::PAYPAL_ORDER_IDENTIFIER_MISSING will be throw before ;)

PR #438, opened 2020-07-06
Install ps_accounts module at installation

v4lux:
Force installation of the module ps_accounts when checkout is installed.


ps_checkout.php
```diff
@@ -138,6 +138,7 @@ public function install()
     {
         // Install for both 1.7 and 1.6
         $defaultInstall = parent::install() &&
+            (new PrestaShop\AccountsAuth\Installer\Install())->installPsAccounts() &&
```
Quetzacoalt91, line 141: Regarding the comments from PHPStan, it seems the PHP library needs to be added.

v4lux: What do you mean by adding the library? Add it via making a composer install before execute phpstan?

Quetzacoalt91: This may require a `composer require` to be done, with the composer.json & composer.lock commited.

PR #443, opened 2020-07-21
Update event condition for building artifact

Quetzacoalt91:
This PR allows artifact to be build on all branches, not only master.

It still takes care to NOT upload it on the release draft until we push to the `branch` master.


Quetzacoalt91:
Yep

PR #444, opened 2020-07-24
Remove 'disabled' attribute from payment confirmation button at check…

pixelicous:
This PR aims to fix an issue where the payment approval button stays greyed out when using paypal express checkout.
Not sure if this is the right way to go about it, but it fixed my problems.

### DIFF
--- a/views/templates/hook/displayPaymentTop.tpl
+++ b/views/templates/hook/displayPaymentTop.tpl
@@ -43,6 +43,8 @@
 
       if ('ps_checkout_expressCheckout' === paymentOptionName) {
         paymentOption.click();
+        const paymentConfirmationButton = document.querySelector('#payment-confirmation button');
+        paymentConfirmationButton.removeAttribute('disabled');
       } else {
         paymentOption.disabled = true;
         paymentOptionContainer.style.display = 'none';


Matt75:
Hi @pixelicous 

Thanks for contribution, it appears this is not the right method to do that.
PayPal SDK have method to enable/disable button. https://developer.paypal.com/docs/business/checkout/add-capabilities/buyer-experience/#synchronous-validation

Can you provide a test scenario to reproduce your issue ? Screenshot or video record ?

Button should be enabled when Terms & Conditions are checked. We know some people disabled it in PrestaShop configuration and currently module do not handle this case. Maybe it's same for you ?

Well this part is totally rewritten in next version to reuse native Terms & Conditions integration and avoid troubles.

If you need assistance, please contact our Support Team thanks to subtab Help on module configuration page.

Have a nice day

pixelicous:
> Hi @pixelicous 
> 
> Thanks for contribution, it appears this is not the right method to do that.
> PayPal SDK have method to enable/disable button. https://developer.paypal.com/docs/business/checkout/add-capabilities/buyer-experience/#synchronous-validation
> 
> Can you provide a test scenario to reproduce your issue ? Screenshot or video record ?
> 
> Button should be enabled when Terms & Conditions are checked. We know some people disabled it in PrestaShop configuration and currently module do not handle this case. Maybe it's same for you ?
> 
> Well this part is totally rewritten in next version to reuse native Terms & Conditions integration and avoid troubles.
> 
> If you need assistance, please contact our Support Team thanks to subtab Help on module configuration page.
> 
> Have a nice day

Hi,

Thank you for the prompt response. 
I don't have much to share, besides my code in my environment is already different than the one presented here due to trying to fix those problems, so can't revert at the moment. 

Out of the box, express checkout doesn't work. Bring up a test lab you can easily see it. 

I'm aware of the terms and conditions part, however this part is not honored during express checkout with this module. 
I'm actually fixing that part right now.. 

There is no radio button after coming back from PayPal to approve those terms and the order placement button stays disabled.

Looking at the code, there is no integration of such check during express checkout.

Talking about implementation, the JS that is incorporated inline right now on the checkout page is also the wrong way of handling what its trying to accomplish. IMHO

Matt75:
Dear @pixelicous 

> Out of the box, express checkout doesn't work. Bring up a test lab you can easily see it.
All release of this module is tested by a dedicated Quality Assurance team before publishing.
Due to various use case (language, currency, theme, configuration) some test scenario cannot be checked but standard tests have been made and validated.
That's why we will introduce more automated tests during this year.

> I'm aware of the terms and conditions part, however this part is not honored during express checkout with this module.
> I'm actually fixing that part right now..
In fact Express Checkout button doesn't require Terms & Conditions because it's only a pre-approval of payment, customer will redirected to order page to choose carrier option, PayPal Order will be updated with additional shipping cost if needed. Customer must approve Terms & Conditions on payment step, that's why button are disabled till customer checked it.

> There is no radio button after coming back from PayPal to approve those terms and the order placement button stays disabled.
This behavior is not right, what is your theme ?

pixelicous:
@Matt75 
I'm not sure i understand, so should there be or shouldn't be a radio button to approve terms.
We are speaking after you get back to the site from paypal, fill in address and pick shipping method

> In fact Express Checkout button doesn't require Terms & Conditions because it's only a pre-approval of payment, customer will redirected to order page to choose carrier option, PayPal Order will be updated with additional shipping cost if needed. Customer must approve Terms & Conditions on payment step << **This is the point where i do not have terms and conditions to approve..**

Modified my files back to show (kind of what my screenshot fo the source showed)
image

Im using a theme called apparel, I don't think it's an issue with the theme, all the building blocks are from core functionality and the theme doesn't override anything.

Matt75:
> I'm not sure i understand, so should there be or shouldn't be a radio button to approve terms.
> We are speaking after you get back to the site from paypal, fill in address and pick shipping method

Yes you should see Terms & Conditions on payment step after customer back to shop to choose shipping method.

> image
> 
> Im using a theme called apparel, I don't think it's an issue with the theme, all the building blocks are from core functionality and the theme doesn't override anything.

This is not right behavior, this is your issue.

pixelicous:
@Matt75 Switching back to classic theme i get the same issue:
image

I do not use any third party modules at all. Reverted back this module to original state and cleared all cache.
Only customisation is language translations in the GUI

I removed any infrastructure security settings.. even though there are no logs anywhere stating i have issues (all of my logs aggregated to kibana so easy for me to find problems)

Matt75:
> @Matt75 Switching back to classic theme i get the same issue:
> image
> 
> I do not use any third party modules at all. Reverted back this module to original state and cleared all cache.
> Only customisation is language translations in the GUI
> 
> I removed any infrastructure security settings.. even though there are no logs anywhere stating i have issues (all of my logs aggregated to kibana so easy for me to find problems)

It should looks like that :
Express Checkout

So please check you use the latest version, clear cache, check you do not have module tpl override inside your theme (if yes update it)

pixelicous:
No overrides, competely new installation, at some point now i even disabled cache.
Anyhow.. I'm out to go fix the code.
Thanks for the support, will be on the lookout to see that 2.0 coming out.

pixelicous:
Issues are now solved by my own code, as described at - https://www.prestashop.com/forums/topic/1028441-new-pspaypal-checkout-module-issues/?tab=comments#comment-3253419

Thanks again.

PR #447, opened 2020-08-03
Move namespaced classes to src

Matt75:


composer.json
```diff
@@ -36,12 +36,12 @@
...
     "psr-4": {
-      "PrestaShop\\Module\\PrestashopCheckout\\": "classes/"
+      "PrestaShop\\Module\\PrestashopCheckout\\": "src"
     },
     "classmap": [
       "ps_checkout.php",
-      "controllers/",
-      "classes/ObjectModel/"
+      "controllers",
+      "classes"
```
yohanBarilleau, line 44: classes folder is still present ?

yohanBarilleau: stay present because the class OrderMatrice has no namespace due to Prestashop

Matt75: ObjectModel cannot be in a namespace because it break internal method of PrestaShop.
We will add a comment in this class to explain it 👍

PR #457, opened 2020-08-19
Sentry

yohanBarilleau:


src/Sentry/SentryProcessor.php
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * @param array $record
+     *
+     * @return array
+     */
+    public function __invoke(array $record)
+    {
+        // record the current user
+        $user = \Context::getContext()->employee;
+        $record['context']['user'] = [
```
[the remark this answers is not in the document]
yohanBarilleau: Fix in next commit ;)

src/Sentry/SentryProcessor.php
```diff
@@ -0,0 +1,50 @@
...
+{
+    /**
+     * @param array $record
+     *
+     * @return array
+     */
+    public function __invoke(array $record)
+    {
+        // record the current user
+        if (\Context::getContext()) {
```
zalexki, line 38: why ? is it even possible for it to be computed as false by PHP ?

src/Sentry/SentryProcessor.php
```diff
@@ -0,0 +1,50 @@
...
+            $user = \Context::getContext()->employee;
+            $record['context']['user'] = [
```
zalexki, line 40: still using unchecked keys is it properly checked somewhere else ?

src/Sentry/SentryProcessor.php
```diff
@@ -0,0 +1,51 @@
...
+     *
+     * @return array
+     */
+    public function __invoke(array $record)
+    {
+        if (!isset($record['context'])) {
+            $record['context'] = [];
+        }
+
+        $user = \Context::getContext()->employee;
```
Matt75, line 41: If you are in front office, employee is null
You should check before use object properties to avoid error

Yo be sure :
- In FO we get email from customer if exist
- In BO we get email from employee
?

Matt75 (changes requested):

src/Sentry/SentryProcessor.php
yohanBarilleau: Yes, Alexis told me the same, i will change it

yohanBarilleau: check in next commit

Matt75:
Are we sure in sentry user context we shouldn't send PrestaShop Account e-mail ? Merchant email used for connect to PrestaShop Account 🤔

Matt75 (approved):

PR #459, opened 2020-08-26
Refacto ajax logout actions

yohanBarilleau:


Matt75 (approved):

yohanBarilleau:
WARNING : in the PSAccount constructor, the idtoken or refreshtoken are empty, it raise an error, so why we reset it ?

PR #462, opened 2020-08-27
Refacto firebase process

yohanBarilleau:


src/Api/Firebase/AuthFactory.php
```diff
@@ -0,0 +1,107 @@
...
+    {
+        return $this->auth->sendPasswordResetEmail($email);
+    }
+
+    /**
+     * @param $response
+     *
+     * @throws \PrestaShop\Module\PrestashopCheckout\Exception\PsCheckoutException
+     */
+    private function savePsAccount($response)
```
Quetzacoalt91, line 96: Looks like we can specify that we expect an array here

yohanBarilleau: added

Quetzacoalt91 (approved):

PR #466, opened 2020-08-27
Segment

yohanBarilleau:


src/ShopUuidManager.php
```diff
@@ -106,4 +106,60 @@ public function isSetForShop($idShop)
 
         return (bool) $this->getForShop($idShop);
     }
+
+    /**
+     * Get one Shop Url
```
Matt75, line 111: Goal of this class is only to manage ShopUuid, I thinks we need a ShopProvider or something like that to retrieve url, protocol etc

yohanBarilleau: Moved to ShopProvider class ;)

Matt75 (approved):

PR #468, opened 2020-08-28
refacto rounding settings

yohanBarilleau:


src/OnBoarding/Step/LiveStep.php
```diff
@@ -39,14 +39,16 @@ public function __construct(PrestaShopConfiguration $configuration)
     /**
      * @param bool $confirmed
      *
-     * @return bool
+     * return bool
```
Quetzacoalt91, line 42:
```suggestion
     * @return bool
```

src/RoundingSettings/RoundingSettings.php
```diff
@@ -0,0 +1,30 @@
...
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@prestashop.com so we can send you a copy immediately.
+ *
+ * @author    PrestaShop SA and Contributors <contact@prestashop.com>
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0  Academic Free License (AFL 3.0)
+ */
+
+namespace PrestaShop\Module\PrestashopCheckout\RoundingSettings;
```
Quetzacoalt91, line 21: Isn't this namespace too restrictive regarding its content? It seems this is the only class we will able to store in it, while a name a little bit more generic like `Settings` would allow you to add more classes in this namespace in the future.

yohanBarilleau: Ok ;)

Quetzacoalt91 (approved):

PR #469, opened 2020-08-28
refacto express checkout

yohanBarilleau:


src/ExpressCheckout/ExpressCheckoutConfiguration.php
```diff
@@ -51,4 +49,34 @@ public function getProductPage()
     {
         return (bool) $this->configuration->get(ExpressCheckout::PS_CHECKOUT_EC_PRODUCT_PAGE);
     }
+
+    /**
+     * @param bool $status
+     *
+     * @throws PsCheckoutException
+     */
+    public function setProductPage($status)
```
Quetzacoalt91, line 58: What do you think about adding `Status` at the end of each method of this class? By looking at the whole diff of this PR, I did not understand why `getProductPage()` returns a boolean only by looking at the method name.

yohanBarilleau: yes, good point !

PR #474, opened 2020-09-01
Fix container error

yohanBarilleau:
To discuss

### DIFF
--- a/controllers/admin/AdminPaypalOnboardingPrestashopCheckoutController.php
+++ b/controllers/admin/AdminPaypalOnboardingPrestashopCheckoutController.php
@@ -23,8 +23,14 @@ use PrestaShop\Module\PrestashopCheckout\Updater\PaypalAccountUpdater;
 
 class AdminPaypalOnboardingPrestashopCheckoutController extends ModuleAdminController
 {
+    /**
+     * @var Ps_checkout
+     */
+    public $module;
+
     public function init()
     {
+        parent::init();
         $idMerchant = Tools::getValue('merchantIdInPayPal');
 
         if (true === empty($idMerchant)) {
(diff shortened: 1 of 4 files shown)


controllers/admin/AdminPaypalOnboardingPrestashopCheckoutController.php
```diff
@@ -23,8 +23,14 @@
...
 class AdminPaypalOnboardingPrestashopCheckoutController extends ModuleAdminController
 {
+    /**
+     * @var Ps_checkout
+     */
+    public $module;
+
     public function init()
     {
+        parent::init();
```
yohanBarilleau, line 33: Why we don't call init parent function before ? Because without this, we can't use the service container from the core

ps_checkout.php
```diff
@@ -81,18 +75,18 @@ class Ps_checkout extends PaymentModule
     ];
 
     public $configurationList = [
-        PayPalConfiguration::INTENT => Intent::CAPTURE,
-        PayPalConfiguration::PAYMENT_MODE => Mode::LIVE,
+        'PS_CHECKOUT_INTENT' => 'CAPTURE',
+        'PS_CHECKOUT_MODE' => 'LIVE',
```
yohanBarilleau, line 79: We should replace all the constants from new classes in the module constructor and install functions AND in the upgrade calls whereas we have an ClassNotFoundException raise when we upgrade the module for the first time.

controllers/admin/AdminPaypalOnboardingPrestashopCheckoutController.php
yohanBarilleau: https://github.com/PrestaShop/PrestaShop/blob/develop/classes/controller/Controller.php#L169 https://github.com/PrestaShop/PrestaShop/blob/develop/classes/controller/AdminController.php#L2769 are the parents' calls

Matt75 (approved):

PR #476, opened 2020-09-03
Switch authorize capture

yohanBarilleau:


_dev/js/back/src/components/panel/payment-mode.vue
```diff
@@ -33,11 +33,63 @@
           <b-form-radio-group
             id="intent-mode"
             v-model="captureMode"
-            :options="intentOptions"
             buttons
             button-variant="outline-primary"
             name="radio-btn-outline"
+          >
+            <b-form-radio value="CAPTURE" v-if="this.captureMode !== 'CAPTURE'" data-toggle="modal" data-target="#modalSwitchAuthorize">
```
Rzial, line 40: `this` can be omitted

_dev/js/back/src/components/panel/payment-mode.vue
```diff
@@ -114,18 +160,27 @@
...
+            : '';
+        if (captureMode !== '') {
+          this.$store.dispatch('updateCaptureMode', captureMode);
+        }
       }
     },
     computed: {
       captureMode: {
         get() {
           return this.$store.state.configuration.captureMode;
```
Rzial, line 183: probably this would be nicer if refactored as a [Vuex getter](https://vuex.vuejs.org/fr/guide/getters.html) and then mapped using `mapGetters`

_dev/js/back/src/components/panel/payment-mode.vue
```diff
@@ -114,18 +160,27 @@
...
+      },
+      updateCaptureMode() {
+        var captureMode =
+          this.captureMode === 'CAPTURE'
+            ? 'AUTHORIZE'
+            : this.captureMode === 'AUTHORIZE'
+            ? 'CAPTURE'
+            : '';
+        if (captureMode !== '') {
+          this.$store.dispatch('updateCaptureMode', captureMode);
```
Rzial, line 176: There is a good practice arround Actions and Mutations:

Define a map of posible actions, mutations
```
export const mutations = {
   SET_FOO = 'SET_FOO',
   SET_BAR = 'SET_BAR',
   SET_BAZ = 'SET_BAZ',
}
```
then on your store:
```
import {mutations} from '<my_fancy_mutations_location>';

const store = {
  mutations: {
    // Even you can export just the symbols to do [SET_FOO] if you dont like the "mutations." part
    [mutations.SET_FOO]: functions () {}
  }
}
```
and finally

```
this.$store.commit(mutations.SET_FOO, payload); // or just SET_FOO as stated before
```

this practice avoid messing up with typos on commits / actions names

PR #482, opened 2020-09-18
refacto PaypalAccountRepository as service

yohanBarilleau:


src/Context/PrestaShopContext.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        $this->context = \Context::getContext();
+    }
+
+    /**
+     * Get the isoCode from the context language, if null, send 'en' as default value
+     *
+     * @return string
+     */
+    public function getIsoCode()
```
Matt75, line 43: Should be named with Lang or Language keyword because we have iso code for country, language, currency...

Matt75 (changes requested):
Please use a variable with cast typing when you use getService to help phpstan and IDE to detect Object used.

For example :
```
/** @var PrestaShop\Module\PrestashopCheckout\Updater\PaypalAccountUpdater $paypalAccountUpdater **/
$paypalAccountUpdater = $this->module->getService('ps_checkout.updater.paypal.account')
$paypalAccountUpdater->update($paypalAccount->getOnboardedAccount());
```

Matt75 (approved):

PR #503, opened 2020-10-02
Load assets only on required page, add version

Matt75:
Load assets on required page only and add version (use of remote attribute is required in this case to work)


NeOMakinG (approved):
lgtm

yohanBarilleau (approved):

PR #520, opened 2020-10-06
fix paypal express checkout button display in each page

yohanBarilleau:


Matt75:
I thinks we should not do that because this is applicable for classic theme but others themes can be different.
So this should be in module css file to allow theme developer and merchants to override it in their theme.
https://devdocs.prestashop.com/1.7/themes/reference/overriding-modules/

views/css/payments.css
```diff
@@ -20,6 +20,12 @@
   margin-top: .55rem;
 }
 
+.product-add-to-cart #ps-checkout-express-button div.paypal-buttons-context-iframe {
+  margin-top:1.25rem;
+  width: 45%;
+  margin-left:75px;
```
NeOMakinG, line 26:
```suggestion
  margin-top: 1.25rem;
  width: 45%;
  margin-left: 75px;
```

Keep consistency inside your css files please, if you don't want to mind about it, you can also configure stylelint to do it for you like inside the PrestaShop main repository

Matt75 (approved):

PR #531, opened 2020-10-08
Avoid duplicate rows with same id_cart in Db

Matt75:


controllers/front/token.php
```diff
@@ -54,19 +54,19 @@ public function postProcess()
...
+
             // If paypal_token_expire is in future, token is not expired
-            if (false !== $psCheckoutCart
-                && (
-                    empty($psCheckoutCart->paypal_token_expire)
-                    || strtotime($psCheckoutCart->paypal_token_expire) <= time()
-                )
+            if (empty($psCheckoutCart->paypal_token_expire)
+                || strtotime($psCheckoutCart->paypal_token_expire) <= time()
             ) {
```
yohanBarilleau, line 65: if useless here because we make the same things in both case

yohanBarilleau (changes requested):

yohanBarilleau (approved):

PR #559, opened 2020-10-20
Fix do not track segment events appends on PrestaShop Installer

Matt75:
We receive too much events due to PrestaShop Installer

### DIFF
--- a/ps_checkout.php
+++ b/ps_checkout.php
@@ -157,7 +157,10 @@ class Ps_checkout extends PaymentModule
      */
     public function install()
     {
+        // When PrestaShop install a module, enable() and install() are called but we want to track only install()
+        // Should be done before parent::install() because enable() will be called first
         $this->disableSegment = true;
+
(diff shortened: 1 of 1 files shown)


Quetzacoalt91:
What do you think about merging these changes inside a common method?

```php
    private function trackModuleAction($action)
    {
        if (defined('_PS_ADMIN_DIR_')) {
            /** @var \PrestaShop\Module\PrestashopCheckout\Segment\SegmentTracker $tracker */
            $tracker = $this->getService('ps_checkout.segment.tracker');
            $tracker->track('Install');
        }
    }
```

This would make the evolution of the check easier as would have to do it only one time

```php
        $this->trackModuleAction('Install');
```

```php
        if ($isEnabled && false === $this->disableSegment) {
            $this->trackModuleAction('Activate');
        }
```

Quetzacoalt91 (approved):

Quetzacoalt91 (approved):

PR #563, opened 2020-10-21
Update dependencies

Matt75:
Fix missing YML component for Service Container in PS 1.6


Matt75:
Yml was already present as dependancy but simply not loaded in 1.6 😔

Rzial (approved):

PR #573, opened 2020-10-22
Update services list for compliancy with SF 2 and 3

Quetzacoalt91:
From PS 1.7.0 to PS 1.7.3, the bundled version of Symfony is 2.x on which the `_defaults` key is invalid. To prevent error on these versions, each service has to be specificaly declared as public.


Quetzacoalt91:
Closed in favor of #574

PR #588, opened 2020-10-27
Do not display CB Inline by PayPal if Hosted Fields are enabled

Matt75:
If Hosted Fields are available in BO and enabled but PayPal SDK Eligibility checker decline, do not display PayPal CB Inline.


Matt75:
Closed in favor of https://github.com/PrestaShopCorp/ps_checkout/pull/590

PR #719, opened 2021-03-01
new ENV loading and sentry exception handling

L3RAZ:


seiwan (approved):

L3RAZ:
!!! Matt's "sentry" branch needs to be tested and merged before mine !!!

Matt75 (dismissed):

seiwan (dismissed):

seiwan (dismissed):

controllers/front/validate.php
```diff
@@ -365,29 +370,29 @@ private function handleException(Exception $exception)
...
                 ],
+                'exceptionCode' => $exception->getCode(),
+                'exceptionMessage' => $exception->getMessage(),
             ],
-            'exceptionCode' => $exception->getCode(),
-            'exceptionMessage' => $exception->getMessage(),
-        ]);
-        exit;
+            500,
+            'Internal Server Error');
```
Quetzacoalt91, line 386: I see that you kept the previous behavior, but having two time the HTTP status + managing the related HTTP code message by yourselves will lead to errors sooner or later.

You may for instance:
* Send HTTP 200 with `Internal Server Error`
* Send the key `httpCode` as 500 but with the second parameter as 404.

I would recommend to stick to only one parameter, i.e having the httpCode in your array, and leave the management of the headers inside your parent class.

If you don't have custom HTTP response code, you could use [`http_response_code()`](https://www.php.net/manual/en/function.http-response-code.php)

Matt75: You're right 👍

Quetzacoalt91 (approved):

seiwan (approved):

PR #721, opened 2021-03-03
[PAYSHIP-654] Send frontend errors to Sentry

Rzial:


src/OnBoarding/Step/LiveStep.php
```diff
@@ -53,4 +54,22 @@ public function isConfirmed()
     {
         return (bool) $this->configuration->get(static::CONFIG_LIVE_STEP);
     }
+
+    /**
+     * @param bool $viewed
```
seiwan, line 59: Missing method description
Missing @return void

_dev/js/back/src/core/plugins/segment.js
```diff
@@ -0,0 +1,32 @@
...
+ * @copyright Since 2007 PrestaShop SA and Contributors
+ * @license   https://opensource.org/licenses/AFL-3.0 Academic Free License 3.0 (AFL-3.0)
+ */
+import Segment from '@prestashopcorp/segment-vue';
+
+export default {
+  install(Vue, { router }) {
+    Vue.use(Segment, {
+      id:
+        process.env.VUE_APP_SEGMENT_API_KEY ||
```
seiwan, line 25: I checked the API key on our Segment app, this one hard coded is the good one and I have also the same on my .env file. Don't you have the same, isn't it ?

Rzial: The hardcoded variable will be removed

seiwan (approved):

PR #725, opened 2021-03-05
CCV Tooltip + Payment Method Logos

Rzial:


seiwan (approved):

ps_checkout.php
```diff
@@ -403,20 +420,99 @@ public function hookDisplayPersonalInformationTop()
...
-     * Express checkout on the cart page
+     * Express checkout and payment method logo block on the cart page
      */
     public function hookDisplayExpressCheckout()
     {
         /** @var \PrestaShop\Module\PrestashopCheckout\PayPal\PayPalPayIn4XConfiguration $payIn4XService */
         $payIn4XService = $this->getService('ps_checkout.pay_in_4x.configuration');
 
+        /** @var \PrestaShop\Module\PrestashopCheckout\FundingSource\FundingSourceProvider $fundingSourceProvider */
+        $fundingSourceProvider = $this->getService('ps_checkout.funding_source.provider');
```
L3RAZ, line 431: My only notice is that this code and code in hookDisplayAdditionalInfo are identical, maybe it could be moved to separate function to avoid duplicate code? :thinking:

PR #736, opened 2021-03-26
[PAYSHIP-302] Fixed line that prevents module from being disabled when uninstalling

L3RAZ:


seiwan (commented):
It's working for disable the module on every shop but we can't reset separately. See more details in the Jira ticket

Matt75:
Unfortunately reset method doesn't work on PrestaShop 1.7 : https://github.com/PrestaShop/PrestaShop/issues/23961
I made severals tests, it work fine on PrestaShop 1.6 but method is never called with PrestaShop 1.7
I tried to find a way to identify reset action on install() and uninstall() but there no useful flag in request for it...

```
ps_checkout.DEBUG: Uninstall module attempted {"request_data":{"_token":"ADMINTOKEN"},"function_args":[],"context_shop_list":[1],"request_symfony":{"_token":"ADMINTOKEN"}} []
ps_checkout.DEBUG: Install module attempted {"request_data":{"_token":"ADMINTOKEN"},"function_args":[],"context_shop_list":[1],"request_symfony":{"_token":"ADMINTOKEN"}} []
```

So it seems we can't fix that for PrestaShop 1.7

PR #739, opened 2021-04-01
Release v2.12.1

seiwan:
- Fix difference between PayPal order and cart amounts


Matt75 (dismissed):

L3RAZ (commented):
Looks good, but I don't like the part where \Context::getContext() is called. It's called in two places (line 100 and 102) when it could be initialized in __contruct()

Matt75 (approved):

PR #760, opened 2021-04-30
Update and rename translations/es.php to traducciones/es.php

alfonsolr09:
UPDATED SPANISH TRANSLATIONS

I don't know why the modified file does not have the same order as the original file.
can you check that it works correctly?
the translations are correct since I am from Spain


alfonsolr09:
I don't know why some translations were duplicated

PR #769, opened 2021-05-25
[PAYSHIP-1129] Added ob_start function for clearing output buffer

L3RAZ:


Matt75:
@L3RAZ This is already performed by Core : https://github.com/PrestaShop/PrestaShop/blob/develop/classes/controller/FrontController.php#L329

L3RAZ:
Ok, then this PR is redundant.

PR #770, opened 2021-05-25
Improve sentry reports

Matt75:
* Change reports level to error
* Exclude reports from vendor directory
* Exclude reports from localhost
* Exclude reports from PHP >= 7.4

### DIFF
--- a/ps_checkout.php
+++ b/ps_checkout.php
@@ -1567,7 +1567,8 @@ class Ps_checkout extends PaymentModule
             $this->sentryClient = new PrestaShop\Module\PrestashopCheckout\Handler\ModuleFilteredRavenClient(
                 $env['PS_CHECKOUT_SENTRY_DSN_MODULE'],
                 [
-                    'level' => 'warning',
+                    'level' => 'error',
+                    'error_types' => E_ERROR,
                     'tags' => [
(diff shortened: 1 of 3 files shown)


seiwan (dismissed):

ps_checkout.php
```diff
@@ -1577,6 +1578,14 @@ private function installSentryExceptionLogger()
             );
 
             $this->sentryClient->setAppPath(realpath(_PS_MODULE_DIR_ . 'ps_checkout/'));
+            $this->sentryClient->setExcludedAppPaths([
+                realpath(_PS_MODULE_DIR_ . 'ps_checkout/vendor/'),
+            ]);
+            $this->sentryClient->setExcludedDomains(['127.0.0.1', 'localhost', '.local']);
+
+            if (version_compare(phpversion(), '7.4.0', '>=')) {
```
Quetzacoalt91, line 1586: It would be interesting that later versions of PrestaShop compatible with PHP 7.4 send errors to sentry again without having to come back on this code?

```suggestion
            if (version_compare(phpversion(), '7.4.0', '>=') && version_compare(_PS_VERSION_, '1.7.8.0', '<')) {
```

seiwan (approved):

PR #774, opened 2021-05-26
Fix guest checkout verification

Flo-180:
Hey,

Prestashop checkout doesn't work with guest checkout.
it's a fix


Matt75:
Closed in favor of https://github.com/PrestaShopCorp/ps_checkout/pull/779
Will be released soon with v2.15.4

PR #821, opened 2021-09-28
Order page rework

L3RAZ:


controllers/admin/AdminAjaxPrestashopCheckoutController.php
```diff
@@ -461,6 +461,8 @@ public function ajaxProcessFetchOrder()
             'moduleName' => $this->module->displayName,
             'orderPayPal' => $presenter->present(),
             'orderPayPalBaseUrl' => $this->context->link->getAdminLink('AdminAjaxPrestashopCheckout'),
+            'moduleLogoUri' => $this->module->getPathUri() . 'logo.png',
+            'orderPaymentLogoUri' => $this->module->getPathUri() . 'views/img/' . strtolower($order->payment) . '.svg',
```
L3RAZ, line 465: Not sure if this is the right way to select payment icon

src/Presenter/Order/OrderPresenter.php
```diff
@@ -225,4 +234,102 @@ private function getTransactionType($type)
             'class' => $class,
         ];
     }
+
+    private function getTotal()
```
L3RAZ, line 238: Would this calculate order total correctly? I couldn't manage to test with few transactions with different currecies.

views/templates/admin/ajaxPayPalOrder.tpl
```diff
@@ -16,161 +16,218 @@
...
+                  <dd>- {$orderPayPalTransaction.paypal_fee} {$orderPayPalTransaction.currency}</dd>
+                  <dt>{l s='Net amount' mod='ps_checkout'}</dt>
+                  <dd>{$orderPayPalTransaction.net_amount} {$orderPayPalTransaction.currency}</dd>
+                </dl>
+              </div>
+              <a href="https://www.paypal.com/activity/payment/{$orderPayPalTransaction.id|escape:'html':'UTF-8'}" target="_blank" class="tabpanel__cta">
+                {l s='See on PayPal' mod='ps_checkout'}
+              </a>
+                {if $orderPayPalTransaction.isRefundable}
+                  <button type="button" class="btn btn-primary btn-sm refund" data-transaction-id="{$orderPayPalTransaction.id|escape:'html':'UTF-8'}">
```
L3RAZ, line 115: Change this button position after mock is updated

controllers/admin/AdminAjaxPrestashopCheckoutController.php
Matt75: PayPal recommends to use https://github.com/paypal/paypal-sdk-logos
But we needs to use webpack like we do for front.js
It means create a third JS app... 🤯
So I think we can use your solution to stay simple

src/Presenter/Order/OrderPresenter.php
Matt75: It should be tested with different currencies, it was designed only for EUR... 😅

Matt75 (approved):

PR #829, opened 2021-11-24
Avoid warning open_basedir restriction

danidomen:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Description?  |  Avoided the "Warning: file_exists(): open_basedir restriction in effect. File(/views/templates/front/XXXXXX.tpl) is not within the allowed path(s)" due a slash before the "views" on the file route. 
| Type?         | bug fix
| BC breaks?    | no
| Deprecations? |  no
| Fixed ticket? | 
| How to test?  | Enable debug mode, check that warning not rise

image

Seems the same problem that on this another module https://github.com/PrestaShop/psgdpr/pull/152

### DIFF
--- a/ps_checkout.php
+++ b/ps_checkout.php
@@ -425,7 +425,7 @@ class Ps_checkout extends PaymentModule
      */
     public function hookDisplayPersonalInformationTop()
     {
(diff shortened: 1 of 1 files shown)


Matt75 (approved):

Matt75:
Hi @danidomen
Thanks a lot for your contribution 👍

PR #835, opened 2021-12-27
Enhance OrderState checking at install

Matt75:
Check if OrderState stored in Configuration exist and come from our module

### DIFF
--- a/src/OrderStates.php
+++ b/src/OrderStates.php
@@ -82,7 +82,7 @@ class OrderStates
         $stateId = (int) \Configuration::getGlobalValue($state);
 
         // Is state ID already existing in the Configuration table ?
-        if (0 === $stateId || false === \OrderState::existsInDatabase($stateId, self::ORDER_STATE_TABLE)) {
+        if (0 === $stateId || false === $this->stateAlreadyExists($stateId)) {
             return $this->createPaypalStateId($state, $color);
         }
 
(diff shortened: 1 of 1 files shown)


src/OrderStates.php
```diff
@@ -237,4 +237,22 @@ private function setStateIcons($state, $orderStateId)
...
+     *
+     * @param int $orderStateId
+     *
+     * @return bool
+     */
+    private function stateAlreadyExists($orderStateId)
+    {
+        $query = new \DbQuery();
+        $query->select('id_order_state');
+        $query->from(static::ORDER_STATE_TABLE);
```
L3RAZ, line 252: Maybe self:: should be used instead? Since it's used everywhere currently

src/OrderStates.php
```diff
@@ -237,4 +237,22 @@ private function setStateIcons($state, $orderStateId)
...
+        $query->where('id_order_state = ' . (int) $orderStateId);
+        $query->where('module_name = "' . static::MODULE_NAME . '"');
```
L3RAZ, line 254: Same here for self::

src/OrderStates.php
Matt75: done

src/OrderStates.php
Matt75: done

L3RAZ (approved):

PR #840, opened 2022-01-11
[PAYSHIP-1] Added error message to display that some correncies are misconfigured

L3RAZ:
Added check for currencies that don't support decimals.

### DIFF
--- a/src/Presenter/Store/Modules/ConfigurationModule.php
+++ b/src/Presenter/Store/Modules/ConfigurationModule.php
@@ -179,7 +179,7 @@ class ConfigurationModule implements PresenterInterface
         return [
             'showError' => !empty($misConfiguredCurrencies),
             'errorMessage' => sprintf(
-                $this->module->l('Attention: you have activated %s currencies, you need to configure those currencies to use 0 decimals as Paypal does not support decimals for those currencies'),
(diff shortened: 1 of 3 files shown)


src/Presenter/Store/Modules/ConfigurationModule.php
```diff
@@ -147,4 +155,33 @@ private function isFundingSourceCardEnabled()
...
+        foreach ($currencies as $currency) {
+            if (in_array($currency['iso_code'], $nonDecimalCurrencies) && (int) $currency['precision'] !== 0) {
+                $misConfiguredCurrencies[] = $currency['iso_code'];
+            }
+        }
+
+        return [
+            'showError' => !empty($misConfiguredCurrencies),
+            'errorMessage' => sprintf(
+                $this->module->l('Attention: you have activated %s currencies, you need to configure those currencies to use 0 decimals as Paypal does not support decimals for those currencies'),
```
Matt75, line 182:
```suggestion
                $this->module->l('Attention: you have activated %s currencies, you need to configure those currencies to use 0 decimals as Paypal does not support decimals for those currencies', 'configurationmodule'),
```
Legacy translations system needs to have filename in lowercase in this case or translations will never be found

Matt75 (changes requested):

Matt75 (dismissed):

PR #844, opened 2022-02-03
Fix ExpressCheckout Address exception

Matt75:
Fix `ExpressCheckoutController - Exception 34 : La propiedad Address->dni está vacía.`
Due to extra field `dni` required in some countries but cannot be filled from PayPal Address


L3RAZ (approved):

controllers/front/ExpressCheckout.php
```diff
@@ -293,4 +306,22 @@ private function addressAlreadyExist($alias, $id_customer)
 
         return (int) Db::getInstance()->getValue($query);
     }
+
+    /**
+     * @param string $state
+     *
+     * @return int|null
+     */
+    private function getStateId($state)
```
L3RAZ, line 315: I would move this function to CountryReporisoty leave as little code in controller as possible

Matt75: Done

seiwan (approved):

PR #852, opened 2022-02-11
Fix unsupported currencies with decimals

seiwan:


src/Presenter/Store/Modules/ConfigurationModule.php
```diff
@@ -166,21 +166,29 @@ private function checkNonDecimalCurrencies()
     {
         $nonDecimalCurrencies = ['HUF', 'JPY', 'TWD'];
 
-        $currencies = \Currency::getCurrencies(false, true);
+        // Enabled currencies for PrestaShop Checkout
+        $query = new \DbQuery();
+        $query->select('id_module');
+        $query->from('module');
+        $query->where('name = "ps_checkout"');
+        $moduleId = \Db::getInstance()->getValue($query);
```
L3RAZ, line 174: I think you are able to use $this->module->id  , but it needs to be checked

L3RAZ (dismissed):

src/Presenter/Store/Modules/ConfigurationModule.php
seiwan: Tested on 1.7 and 1.6. It works. Thanks

PR #853, opened 2022-02-24
[PAYSHIP-1557] Fix ExpressCheckout password hash method

Matt75:


controllers/front/ExpressCheckout.php
```diff
@@ -189,7 +189,16 @@ private function createCustomer($email, $firstName, $lastName)
         $customer->email = $email;
         $customer->firstname = $firstName;
         $customer->lastname = $lastName;
-        $customer->passwd = md5(time() . _COOKIE_KEY_);
+
+        if (class_exists('PrestaShop\PrestaShop\Core\Crypto\Hashing')) {
+            $crypto = new PrestaShop\PrestaShop\Core\Crypto\Hashing();
+            $customer->passwd = $crypto->hash(
```
L3RAZ, line 195: I see that the password is hashed 2 times here, maybe it should be like this?  $crypto->hash(time(), _COOKIE_KEY_)

controllers/front/ExpressCheckout.php
```diff
@@ -189,7 +189,16 @@ private function createCustomer($email, $firstName, $lastName)
...
+                md5(time() . _COOKIE_KEY_),
+                _COOKIE_KEY_
+            );
+        } else {
+            $customer->passwd = md5(time() . _COOKIE_KEY_);
```
L3RAZ, line 200: And in PS passwords are hashed like this md5(_COOKIE_KEY_ . $passwd)   or maybe it doesn't matter :thinking: ?

controllers/front/ExpressCheckout.php
Matt75: That's right

PR #870, opened 2022-04-11
PAYSHIP-1693 Feature/pay4x config refactor

L3RAZ:


Rzial (approved):

upgrade/upgrade-2.18.0.php
```diff
@@ -31,7 +31,7 @@
 function upgrade_module_2_18_0($module)
 {
     /** @var \PrestaShop\Module\PrestashopCheckout\FundingSource\FundingSourceConfigurationRepository $fundingSourceConfiguration */
-    $fundingSourceConfiguration = $module->get('ps_checkout.funding_source.configuration.repository');
+    $fundingSourceConfiguration = $module->getService('ps_checkout.funding_source.configuration.repository');
```
Matt75, line 34: You cannot use Service Container inside upgrade script.
PrestaShop will create an instance of ps_checkout then it will launch the upgrade process by replacing all files of the module by files from ZIP archive.
It means instance of module passed to upgrade script is the previous version with previous service container.
So if merchant upgrade from a version where this service is not available or if this function is not exist, it will throw an error.

Matt75 (changes requested):

upgrade/upgrade-2.18.0.php
seiwan: I made requested modifications

PR #906, opened 2022-07-18
Ensure Address::validateFields return true to continue

Matt75:
ExpressCheckout: Address is not created
Address::validateFields can return bool or string

### DIFF
--- a/controllers/front/ExpressCheckout.php
+++ b/controllers/front/ExpressCheckout.php
@@ -278,7 +278,7 @@ class ps_checkoutExpressCheckoutModuleFrontController extends AbstractFrontContr
             $address->id_state = $idState;
         }
 
-        if ($address->validateFields(false)) {
+        if (true !== $address->validateFields(false)) {
             return false;
         }
 


controllers/front/ExpressCheckout.php
```diff
@@ -278,7 +278,7 @@ private function createAddress(
             $address->id_state = $idState;
         }
 
-        if ($address->validateFields(false)) {
+        if (true !== $address->validateFields(false)) {
```
PululuK, line 281:
```suggestion
        if (!$address->validateFields(false)) {
```

PululuK:
Hi @Matt75 
It is a critical issue, can you please provide a release ?

controllers/front/ExpressCheckout.php
Matt75: Address::validateFields can return bool or string

PululuK: @Matt75  in your case it will never return a string because `error_return` param is false. See `ObjectModel::validateFields` please

Matt75: Yes but we want strictly ensure case with bool

PR #920, opened 2022-08-18
PAYSHIP-1813 Added webhook secret storing and validation

L3RAZ:


controllers/admin/AdminAjaxPrestashopCheckoutController.php
```diff
@@ -1003,4 +1003,25 @@ private function togglePayLaterConfiguration($method)
 
         $this->ajaxDie(json_encode(true));
     }
+
+    private function ajaxProcessUpsertSecretToken()
+    {
+        /** @var \PrestaShop\Module\PrestashopCheckout\Webhook\WebhookSecretTokenService $webhookSecretTokenService */
+        $webhookSecretTokenService = $this->module->getService('  ps_checkout.webhook.service.secret_token');
```
Matt75, line 1010:
```suggestion
        $webhookSecretTokenService = $this->module->getService('ps_checkout.webhook.service.secret_token');
```

controllers/admin/AdminAjaxPrestashopCheckoutController.php
```diff
@@ -1003,4 +1003,25 @@ private function togglePayLaterConfiguration($method)
...
+
+        $secret = (string) Tools::getValue('body');
```
Matt75, line 1012:
```suggestion
        $secret = (string) Tools::getValue('body');
```
Is there a validation performed after ?

L3RAZ: No, I think the only validation I know right now would be to check if it's not empty - I can add that

PR #924, opened 2022-09-09
Don't update addresses if already used

PululuK:


PululuK:
ping @Matt75

Matt75:
This change will generate too much address, this table can be huge already natively, generate a new address for each ExpressCheckout session will increase it furthermore.


It require to store an association between an alias, a checksum of PrestaShop Address and a checksum of PayPal Address.

Checksum of PayPal Address will be used to retrieve PrestaShop Address if exist

PR #925, opened 2022-09-12
Check if country contains states to prevent UK issue

PululuK:


Matt75:
Duplicate of https://github.com/PrestaShopCorp/ps_checkout/pull/918

PR #987, opened 2023-02-28
Add payment fields component for APMs

Matt75:


L3RAZ:
@Matt75 Added tests and I think it's ready for review

PR #1009, opened 2023-03-21
Upgrade PHPStan to the latest version

Quetzacoalt91:
Since the first implementation of PHPStan in our development process, we did not increase the version because the docker image had changed. We were stuck on `PHPStan - PHP Static Analysis Tool 0.12.89`, a version released on 09/06/2021.

Keeping our tools up-to-date is important for the stability of our projects. However newer versions were using PHP 8 in their Docker image, preventing it to work with old version of PrestaShop.
This PR add PHPStan as a composer dependency so we can control the version used with the lock file, and we updated the CI workflow so PrestaShop versions are manually downloaded and unzipped.

This is a work in progress, there is still some work to do:
- [ ] ...


Quetzacoalt91:
> We now have 3 branches dedicated to each major version of PrestaShop:
> 
>     * `prestashop/1.6.1.x`
> 
>     * `prestashop/1.7.x`
> 
>     * `prestashop/8.x`
> 
> 
> Code is common along theses branches, differences are only on dependancies so on composer.json and composer.lock and also on GitHub Actions.

Understood, thanks. At the first glance it was quite difficult to find which one in the 60+ branches was actually active. :sweat_smile: 
These branches specific to PrestaShop versions will make the work easier. I spread the changes as soon I have something good enough in this PR.

PR #1010, opened 2023-03-21
Upgrade PHPStan to latest version (PrestaShop 8)

Quetzacoalt91:
Since the first implementation of PHPStan in our development process, we did not increase the version because the docker image had changed. We were stuck on `PHPStan - PHP Static Analysis Tool 0.12.89`, a version released on 09/06/2021.

Keeping our tools up-to-date is important for the stability of our projects. However newer versions were using PHP 8 in their Docker image, preventing it to work with old version of PrestaShop.
This PR add PHPStan as a composer dependency so we can control the version used with the lock file, and we updated the CI workflow so PrestaShop versions are manually downloaded and unzipped.

- [ ] Some errors will appear with the rules added. There are to way to fix this: Fix all the errors in the same PR, or add a baseline. I would prefer the first solution as it may fix some issues.

Replaces #1009

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -56,35 +56,48 @@ jobs:
         uses: actions/checkout@v2.0.0
       - name: Run PHP-CS-Fixer
         uses: prestashopcorp/github-action-php-cs-fixer@master
+
   phpstan:
     name: PHPStan
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


Quetzacoalt91:
@Matt75 I need your opinion on this PR and #1011. How confident are you in fixing errors in this PR? Otherwise we can ignore them until the code is modified for the first time.

PR #1012, opened 2023-03-22
Cannot identify right order status in Split Order

ashok1706:
After the creating split order, module can't change order status second order, it only change first order which store in module->currentOrder.


Matt75:
Hi @ashok1706 

I'm sorry but this change is not a good solution.
Don't worry we will fix it in right way soon.

PR #1043, opened 2023-05-15
Wrong state when using express checkout

RubenHostinato:
Hello, i've edited the code in CountryRepository as the function getStateId could return completely wrong results.
As an example, there's a shop with states in italy, united states, indonesia and india.
If i create an account with expresscheckout using an italian address with state "Varese" which in short is "VA", the address gets created with state "Nevada".
That's because the current query does not take into account the country and it simply search "VA" in the name.
Furthermore, it should search first into the iso_code as there could be issues as well. Taking into account the country "VA" could return "Genova", "Mantova" ecc...
I hope this solves many problems.


Matt75:
Hello @RubenHostinato 

Thanks for your feedback and your contribution, we will try to fix it soon as possible

Matt75:
Merged into https://github.com/PrestaShopCorp/ps_checkout/pull/1060

PR #1206, opened 2024-03-01
Add hookActionOrderSlipAdd

Matt75:


config/common.yml
```diff
@@ -684,7 +686,7 @@ services:
     arguments:
       - "@ps_checkout.repository.pscheckoutcart"
 
-  PrestaShop\Module\PrestashopCheckout\PayPal\Payment\Refund\CommandHandler\RefundPayPalCaptureCommandHandler:
```
L3RAZ, line 686: Why change back the dependency name from full class name to **ps_checkout.command.handler.paypal.payment.refund.refund_paypal_capture** ?
Going forward we should create services/dependencies with full class name, as it's more convenient to call **getService(Service::class)**  than having to remember what the service is named in config

btafforeau: I had a problem with my dependencies, as it was the only one wrote with this syntax i thought that it may be the problem (it was obviously not the problem). Forgot to change it back.
As it's more convenient for the getService calls, I will add the new Query in this new style, as well as going back on this change

matks:
Hello I close this PR :/ sorry but with the refactoring that happened for v5 I think the codebase changed too much. It is maybe faster to redo the PR starting from branch `5.0.x` than trying to fix the git conflicts of this branch

The code will not be lost it will remain available in this PR

PR #1214, opened 2024-03-11
Release v6.3.6.1

Matt75:


LaqueP:
Just updated today the module, We are facing issue on prestashop 1.6
Recommended Card Fields options, didn't work.
Only Paypal card fields works correctly
card fields problem

PR #1231, opened 2024-05-07
[PAYSHIP-2824] GitHub workflow PS8

seiwan:


seiwan:
This PR is ready to review but we have to wait some devops actions (already requested) :
- New GCP accounts + permissions to use the action gcp-auth (workload-federation) for each env
- Add the secrets on WI_PROVIDER and  WI_SA for each env
- Create the GCP bucket storages on integration and preprod (should be called checkout-assets-preproduction and  checkout-assets-integration)

Add the labels **ps8** and **integration deployment** / **preproduction deployment** to trigger the workflow
Labels are not created yet, so you can change them in the PR if you want.

@Matt75 FYI :
I added a condition to trigger PHP and JS tests only when we will add the label **ready to review**
I created the secrets stores on projects **ps-checkout-<environment>**, it's called **module-env**. I cleaned it a little bit in prod (with help from @L3RAZ), but you can take a look.

.github/workflows/ps8-build-release.yml
```diff
@@ -0,0 +1,222 @@
...
+  pull_request:
+    types: [opened, reopened, synchronize, edited, labeled]
+  push:
+    tags:
+      - "v8.*"
+    branches:
+      - "prestashop/8.x"
+
+env:
+  ZIP_NAME: ${{ github.event.repository.name }}-${{ github.ref_name }}
```
elodie-bil3, line 13: Hello, I think that this env variable will override the one that is directly inside the jobs

seiwan: Discussed with you. Zip names for preprod and integration are OK.

Matt75 (dismissed):

Matt75 (approved):

PR #1294, opened 2024-11-29
[PAYSHIP-3121] Apple Pay on PS 1.7

L3RAZ:


CorentinFlament (dismissed):

Matt75 (dismissed):

ps_checkout.php
```diff
@@ -105,16 +106,19 @@ class Ps_checkout extends PaymentModule
...
+        'PS_CHECKOUT_DOMAIN_REGISTERED_LIVE' => false,
     ];
 
     public $confirmUninstall;
     public $bootstrap;
 
     // Needed in order to retrieve the module version easier (in api call headers) than instanciate
     // the module each time to get the version
-    const VERSION = '7.3.6.3';
+    const VERSION = '7.4.2.2';
```
seiwan, line 119: It can be 7.4.2.0, actually, as you prefer.

ps_checkout.php
```diff
@@ -133,7 +137,7 @@ public function __construct()
 
         // We cannot use the const VERSION because the const is not computed by addons marketplace
         // when the zip is uploaded
-        $this->version = '7.3.6.3';
+        $this->version = '7.4.2.2';
```
seiwan, line 140: It can be 7.4.2.0, actually, as you prefer.

config.xml
```diff
@@ -2,7 +2,7 @@
 <module>
     <name>ps_checkout</name>
     <displayName><![CDATA[PrestaShop Checkout]]></displayName>
-    <version><![CDATA[7.3.6.3]]></version>
+    <version><![CDATA[7.4.2.2]]></version>
```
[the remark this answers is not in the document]
L3RAZ: I will keep 4.2.2 Just so we know that capture retries and other fixes are added

ps_checkout.php
L3RAZ: Same here

ps_checkout.php
L3RAZ: And here

seiwan (dismissed):

seiwan:
Approved, but there is some issues on the CI/CD

.github/workflows/checkout-cd-prod-beta.yml
seiwan, line 1: Actually, I think we can delete this workflow.

seiwan (changes requested):

seiwan (approved):

PR #1298, opened 2024-12-04
[PAYSHIP-3029] Added locale and error message translations for FO SDK

L3RAZ:


ps_checkout.php
```diff
@@ -1121,6 +1122,8 @@ public function hookActionFrontControllerSetMedia()
                 'checkout.payment.token.delete.modal.content' => $this->l('The following payment method will be deleted from your account:'),
                 'checkout.payment.token.delete.modal.confirm-button' => $this->l('Delete payment method'),
                 'checkout.payment.loader.processing-request' => $this->l('Please wait, we are processing your request'),
+                'APPLE_PAY_MERCHANT_SESSION_VALIDATION_ERROR' => $this->l('We’re unable to process your Apple Pay payment at the moment. This could be due to an issue verifying the payment setup for this website. Please try again later or choose a different payment method.'),
```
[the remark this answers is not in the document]
L3RAZ: The format doesn't really matter, since the translations are simple key-value pair. This is done because we get this error in Apple Pay and we can just use translation service to display the error as a user-friendly message.

ps_checkout.php
```diff
@@ -1018,6 +1018,7 @@ public function hookActionFrontControllerSetMedia()
 
         Media::addJsDef([
             $this->name . 'Version' => $version->getSemVersion(),
+            $this->name . 'Locale' => $this->context->language->iso_code,
```
Matt75, line 1021: `$this->context->language->iso_code` returns only the language code (e.g., `en`, `fr`), not a full locale code.

The variable name `locale` is currently misnamed as it actually holds a country language code. To improve clarity and accuracy, we should either rename the variable to reflect its current value or change the value format to follow standard locale codes.
 
To comply with standard locale formats and [PayPal requirements](https://developer.paypal.com/sdk/js/configuration/#link-locale), we should set a real locale code.

This can be done by constructing the locale code using both the language and country codes, ensuring it is reusable and compliant with PayPal's standards.

seiwan (dismissed):

ps_checkout.php
L3RAZ: While I would agree with you, this variable is only used for frontoffice-sdk translations and not in PayPal API

Matt75 (approved):

PR #1304, opened 2024-12-13
[PAYSHIP-3149] PrestaShop 9 compatibility

L3RAZ:


src/Order/CommandHandler/UpdateOrderStatusCommandHandler.php
```diff
@@ -80,7 +72,7 @@ public function handle(UpdateOrderStatusCommand $command)
             throw new OrderException(sprintf('Failed to update status or send email when changing OrderState #%d of Order #%d.', $command->getNewOrderStatusId()->getValue(), $command->getOrderId()->getValue()), OrderException::FAILED_UPDATE_ORDER_STATUS);
         }
 
-        $this->eventDispatcher->dispatch(new OrderStatusUpdatedEvent($orderStateId));
+//        $this->eventDispatcher->dispatch(new OrderStatusUpdatedEvent($orderStateId));
```
seiwan, line 75: Maybe we can remove it if it's not used anymore ?

src/Order/Matrice/CommandHandler/UpdateOrderMatriceCommandHandler.php
```diff
@@ -20,33 +20,28 @@
...
         $orderMatrice = new \OrderMatrice();
         $orderMatrice->id_order_prestashop = $command->getOrderId()->getValue();
         $orderMatrice->id_order_paypal = $command->getOrderPayPalId()->getValue();
 
         $res = $orderMatrice->add();
 
-        if (!empty($res)) {
-            $this->eventDispatcher->dispatch(new OrderMatriceUpdatedEvent());
-        }
+//        if (!empty($res)) {
```
seiwan, line 43: Maybe we can remove it if it's not used anymore ?

src/PayPal/PaymentToken/Query/GetCustomerPaymentTokensQueryHandler.php
```diff
@@ -45,7 +45,7 @@ public function __construct(PaymentTokenRepository $paymentTokenRepository)
      *
      * @throws Exception
      */
-    public function handle(GetCustomerPaymentTokensQuery $query)
+    public function __invoke(GetCustomerPaymentTokensQuery $query)
     {
 //        $paymentTokens = $this->paymentTokenRepository->findByPrestaShopCustomerId($query->getCustomerId()->getValue(), $query->getPageSize(), $query->getPageNumber());
```
seiwan, line 50: Maybe we can remove it if it's not used anymore ?

src/Repository/PsCheckoutCartRepository.php
```diff
@@ -42,28 +47,39 @@ public function __construct(CacheInterface $cartPrestaShopCache)
...
+            $psCheckoutCartCollection = new PrestaShopCollection('PsCheckoutCart');
+            $psCheckoutCartCollection->where('id_cart', '=', (int) $cartId);
+            $psCheckoutCartCollection->orderBy('date_upd', 'desc');
+
+            /** @var PsCheckoutCart|false $psCheckoutCart */
+            $psCheckoutCart = $psCheckoutCartCollection->getFirst();
+            return $psCheckoutCart;
+        });
+
+//        if ($this->cartPrestaShopCache->hasItem(CacheSettings::CART_ID . $cartId)) {
```
seiwan, line 67: Maybe we can remove it if it's not used anymore ?

src/Repository/PsCheckoutCartRepository.php
```diff
@@ -42,28 +47,39 @@ public function __construct(CacheInterface $cartPrestaShopCache)
...
-        /** @var \PsCheckoutCart|false $psCheckoutCart */
+        /** @var PsCheckoutCart|false $psCheckoutCart */
         $psCheckoutCart = $psCheckoutCartCollection->getFirst();
 
         if (false !== $psCheckoutCart) {
-            $this->cartPrestaShopCache->setMultiple([
-                CacheSettings::CART_ID . $cartId => $psCheckoutCart,
-                CacheSettings::PAYPAL_ORDER_ID . $psCheckoutCart->paypal_order => $psCheckoutCart,
-            ]);
+//            $this->cartPrestaShopCache->setMultiple([
```
seiwan, line 79: Maybe we can remove it if it's not used anymore ?

src/Repository/PsCheckoutCartRepository.php
```diff
@@ -114,23 +140,30 @@ public function save(\PsCheckoutCart $psCheckoutCart)
...
-                CacheSettings::PAYPAL_ORDER_ID . $psCheckoutCart->paypal_order => $psCheckoutCart,
-            ]);
+            $cacheItem = $this->cartPrestaShopCache->getItem(CacheSettings::CART_ID . $psCheckoutCart->id_cart);
+            $cacheItem->set($psCheckoutCart);
+            $this->cartPrestaShopCache->save($cacheItem);
+            $cacheItem = $this->cartPrestaShopCache->getItem(CacheSettings::PAYPAL_ORDER_ID . $psCheckoutCart->paypal_order);
+            $cacheItem->set($psCheckoutCart);
+            $this->cartPrestaShopCache->save($cacheItem);
+
+//            $this->cartPrestaShopCache->setMultiple([
```
seiwan, line 150: Maybe we can remove it if it's not used anymore ?

controllers/front/webhook.php
```diff
@@ -178,11 +178,11 @@ protected function displayRestrictedCountryPage()
      *
      * @see FrontController::geolocationManagement()
      *
-     * @param Country $defaultCountry
+     * @param Country $default_country
      *
      * @return false
      */
-    protected function geolocationManagement($defaultCountry)
+    protected function geolocationManagement($default_country)
```
kpodemski, line 185: same as above, what's the point of this change?

ps_checkout.php
```diff
@@ -197,22 +170,8 @@ public function install()
...
-
-            return $result;
-        }
-
-        // Install specific to prestashop 1.7
-        if ($shopContext->isShop17()) {
-            $result = $result && (bool) $this->registerHook(self::HOOK_LIST_17);
-            $this->updatePosition(\Hook::getIdByName('paymentOptions'), false, 1);
-        }
+        $this->updatePosition(\Hook::getIdByName('paymentOptions'), false, 1);
```
kpodemski, line 174:
```suggestion
        $this->updatePosition(Hook::getIdByName('paymentOptions'), false, 1);
```

you are not using namespaces here, right?

ps_checkout.php
```diff
@@ -1560,34 +1381,10 @@ public function hookDisplayPaymentByBinaries(array $params)
...
-        if (1 === (int) Configuration::get('PS_ORDER_PROCESS_TYPE')) {
-            return $this->context->link->getPageLink(
-                'order-opc',
-                true,
-                (int) $this->context->language->id
-            );
-        }
-
-        // PrestaShop 1.6 standard checkout
         return $this->context->link->getPageLink(
```
kpodemski, line 1384: It's a tiny detail but `$this->context->link->getPageLink('order')` would've been enough

src/Order/CommandHandler/AddOrderPaymentCommandHandler.php
```diff
@@ -100,6 +96,6 @@ public function handle(AddOrderPaymentCommand $command)
         $orderPayments = $order->getOrderPayments();
         $latestOrderPayment = end($orderPayments);
 
-        $this->eventDispatcher->dispatch(new OrderPaymentCreatedEvent((int) $latestOrderPayment->id));
+//        $this->eventDispatcher->dispatch(new OrderPaymentCreatedEvent((int) $latestOrderPayment->id));
```
kpodemski, line 99: :)

src/PayPal/Order/EventProcessor/PayPalOrderEventProcessor.php
```diff
@@ -0,0 +1,240 @@
...
+    private OrderStateMapper $orderStateMapper;
+    private PayPalConfiguration $payPalConfiguration;
+    private PayPalOrderRepository $payPalOrderRepository;
+    private SavePayPalOrderCommandHandler $savePayPalOrderCommandHandler;
+    private SaveCheckoutCommandHandler $saveCheckoutCommandHandler;
+    private SavePayPalOrderStatusCommandHandler $savePayPalOrderStatusCommandHandler;
+    private QueryBusInterface $queryBus;
+    private UpdateOrderStatusCommandHandler $updateOrderStatusCommandHandler;
+
+    public function __construct(
```
kpodemski, line 67: For PS9 you can use constructor property promotion

src/PayPal/Order/PayPalOrderPresenter.php
```diff
@@ -95,10 +97,7 @@ public function getTotalAmountFormatted()
             return '';
         }
 
-        return Tools::displayPrice(
-            (float) $this->paypalOrderDataProvider->getTotalAmount(),
-            Currency::getCurrencyInstance(Currency::getIdByIsoCode($this->paypalOrderDataProvider->getCurrencyCode()))
-        );
+        return \Context::getContext()->getCurrentLocale()->formatPrice((float) $this->paypalOrderDataProvider->getTotalAmount(), $this->paypalOrderDataProvider->getCurrencyCode());
```
kpodemski, line 100:
```suggestion
        return \Context::getContext()->getCurrentLocale()->formatPrice((float) $this->paypalOrderDataProvider->getTotalAmount(), $this->paypalOrderDataProvider->getCurrencyCode());
```

I think this could be simplified with `Tools::getContextLocale`

src/PaypalCountryCodeMatrice.php
```diff
@@ -64,10 +64,10 @@ public function getPrestashopIsoCode($isoCode)
             return false;
         }
 
-        if (false === array_search($isoCode, self::MATCH_ISO_CODE)) {
+        if (!in_array($isoCode, self::MATCH_ISO_CODE)) {
             return $isoCode;
         }
 
-        return array_search($isoCode, self::MATCH_ISO_CODE);
+        return in_array($isoCode, self::MATCH_ISO_CODE);
```
kpodemski, line 71: Something is off here, I think, because if the purpose is that you return the mapped value if it exists in the static then it will not do that after your change if I'm not wrong.

```
public function getPrestashopIsoCode($isoCode)
{
    if (!is_string($isoCode)) {
        return false;
    }

    $mappedCode = array_search($isoCode, self::MATCH_ISO_CODE);
    if ($mappedCode !== false) {
        return $mappedCode;
    }

    return $isoCode;
}
```

views/templates/hook/displayOrderConfirmation.tpl
```diff
@@ -40,15 +33,15 @@
 
     {if $approvalLink && $orderPayPalStatus === 'PENDING_APPROVAL'}
       <div class="alert alert-warning">
-        <a href="#ps_checkout-displayPaymentReturn" class="alert-link"{if $isShop17} style="margin: initial;padding: initial;"{/if}>
+        <a href="#ps_checkout-displayPaymentReturn" class="alert-link" style="margin: initial;padding: initial;">
             {$translations.notificationPendingApproval|escape:'html':'UTF-8'}
```
kpodemski, line 37:
```suggestion
            {$translations.notificationPendingApproval}
```

using such escapes in front office templates is no longer required

seiwan (changes requested):

config/common.yml
```diff
@@ -42,53 +42,58 @@ services:
...
     class: 'PrestaShop\Module\PrestashopCheckout\CommandBus\TacticianCommandBusAdapter'
     public: true
-    arguments:
-      - "@ps_checkout.tactician.bus"
+    factory: [ '@PrestaShop\Module\PrestashopCheckout\CommandBus\TacticianCommandBusFactory', "create" ]
+
+  ps_checkout.bus.query:
+    class: 'PrestaShop\Module\PrestashopCheckout\CommandBus\TacticianQueryBusAdapter'
+    public: true
+    factory: [ '@PrestaShop\Module\PrestashopCheckout\CommandBus\TacticianQueryBusFactory', "create" ]
```
jolelievre, line 53: In PrestaShop V9 we removed tactician in favor of Symfony Messenger as the middleware that handles the command bus

I'm not sure forcing a custom integration of tactician in this module is a good idea, ideally you should rely on the core messenger like any other CQRS commands The migration is quite straight forward actually you just need to add this attribute on the handler class `#[AsCommandHandler]` and add `autowire` and `autoconfigure` on the handler service definition

config/common.yml
```diff
@@ -42,53 +42,58 @@ services:
...
-        PrestaShop\Module\PrestashopCheckout\PayPal\Customer\Command\SavePayPalCustomerCommand: 'PrestaShop\Module\PrestashopCheckout\PayPal\Customer\CommandHandler\SavePayPalCustomerCommandHandler'
-        PrestaShop\Module\PrestashopCheckout\PayPal\OAuth\Query\GetPayPalGetUserIdTokenQuery: 'PrestaShop\Module\PrestashopCheckout\PayPal\OAuth\Query\GetPayPalGetUserIdTokenQueryHandler'
-        PrestaShop\Module\PrestashopCheckout\PayPal\Order\Command\SavePayPalOrderCommand: 'PrestaShop\Module\PrestashopCheckout\PayPal\Order\CommandHandler\SavePayPalOrderCommandHandler'
-        PrestaShop\Module\PrestashopCheckout\PayPal\GooglePay\Query\GetGooglePayTransactionInfoQuery: 'PrestaShop\Module\PrestashopCheckout\PayPal\GooglePay\Query\GetGooglePayTransactionInfoQueryHandler'
-        PrestaShop\Module\PrestashopCheckout\PayPal\ApplePay\Query\GetApplePayPaymentRequestQuery: 'PrestaShop\Module\PrestashopCheckout\PayPal\ApplePay\Query\GetApplePayPaymentRequestQueryHandler'
+      - PrestaShop\Module\PrestashopCheckout\Order\Command\AddOrderPaymentCommand: ['@PrestaShop\Module\PrestashopCheckout\Order\CommandHandler\AddOrderPaymentCommandHandler']
```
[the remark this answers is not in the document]
jolelievre: I agree (see my comment above) you should relay on the core command bus instead of an internal implementation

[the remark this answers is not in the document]
jolelievre: Also using autowiring on your services (along with explicit type for service injections) should greatly simplify the service definitions

config/common.yml
```diff
@@ -163,12 +187,27 @@ services:
         '@PrestaShop\Module\PrestashopCheckout\PayPal\PaymentToken\EventSubscriber\PaymentMethodTokenEventSubscriber'
       ]
 
+  ps_checkout.event.dispatcher.symfony2:
+    class: 'Symfony\Component\EventDispatcher\EventDispatcherInterface'
+    factory: [ '@PrestaShop\Module\PrestashopCheckout\Event\SymfonyEventDispatcherFactory', "create" ]
```
jolelievre, line 192: Quick question as I don't know the details of this module, but why do you need a specific event dispatcher? What does it do that the regular global event dispatcher from Symfony can't do?

Maybe it's because those services are also used on the FO, in a legacy environment that doesn't include the Symfony event dispatcher?

jolelievre: Another question however, what is the difference between `ps_checkout.event.dispatcher.symfony` and `ps_checkout.event.dispatcher.symfony2` I mean I see that they don't have the same subscribers but why? 

Shouldn't the initial `ps_checkout.event.dispatcher.symfony` be cleaned if not used anymore? or maybe simply modified to match this new need?

config/common.yml
jolelievre: Oh actually I noticed that the CQRS commands are also used in FO controllers, I guess that explains the need for this custom bus right?

src/Checkout/EventSubscriber/CheckoutEventSubscriber.php
```diff
@@ -43,30 +46,27 @@
...
+    private CreateOrderCommandHandler $createOrderCommandHandler;
+    private CapturePayPalOrderCommandHandler $capturePayPalOrderCommandHandler;
+
+    public function __construct(
+        CheckoutChecker $checkoutChecker,
+        PsCheckoutCartRepository $psCheckoutCartRepository,
+        QueryBusInterface $queryBus,
+        UpdatePaymentMethodSelectedCommandHandler $updatePaymentMethodSelectedCommandHandler,
+        CreateOrderCommandHandler $createOrderCommandHandler,
+        CapturePayPalOrderCommandHandler $capturePayPalOrderCommandHandler
```
jolelievre, line 62:
```suggestion
        private readonly CheckoutChecker $checkoutChecker,
        private readonly PsCheckoutCartRepository $psCheckoutCartRepository,
        private readonly QueryBusInterface $queryBus,
        private readonly UpdatePaymentMethodSelectedCommandHandler $updatePaymentMethodSelectedCommandHandler,
        private readonly CreateOrderCommandHandler $createOrderCommandHandler,
        private readonly CapturePayPalOrderCommandHandler $capturePayPalOrderCommandHandler
```
Little tip since you now use PHP 8.1 minimum, you can use the property promotion feature, if you want, that greatly decreases the quantity of code https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion

src/CommandBus/TacticianCommandBusAdapter.php
```diff
@@ -20,19 +20,19 @@
 
 namespace PrestaShop\Module\PrestashopCheckout\CommandBus;
 
-use League\Tactician\CommandBus;
+use PrestaShopBundle\CommandBus\MessengerCommandBus;
```
jolelievre, line 23: Ah seems like you actually do use the `MessengerCommandBus` adpter from the core, my bad I was misled by the name of this class But since it's no longer based on Tactitian maybe it should be renamed as well (along with its factory)?

src/Order/CommandHandler/AddOrderPaymentCommandHandler.php
```diff
@@ -51,15 +45,17 @@ class AddOrderPaymentCommandHandler extends AbstractOrderHandler
...
-        EventDispatcherInterface $eventDispatcher,
         FundingSourceTranslationProvider $fundingSourceTranslationProvider,
         PayPalConfiguration $configuration
     ) {
-        $this->eventDispatcher = $eventDispatcher;
         $this->fundingSourceTranslationProvider = $fundingSourceTranslationProvider;
         $this->configuration = $configuration;
     }
 
+    public function __invoke(AddOrderPaymentCommand $command) {
```
jolelievre, line 55: By the way, if you use the `AsCommandHandler` you shouldn't need to add this `__invoke` method, but again I'm really aware of this module's architecture and context If the command/query buses are used in FO/legacy context this auto-configuration will probably not work

jolelievre (commented):
Thanks @L3RAZ for this huge work

I have a few questions, sorry if they don't make sens it's probably because I lack knowledge on this module's code 😅

src/Order/CommandHandler/AddOrderPaymentCommandHandler.php
L3RAZ: I will check if this works on front controllers

config/common.yml
L3RAZ: I added the second event dispatcher because we have a circular dependency in one command handler. So the first event dispatcher calls CaptureOrderCommandHandler and the second event dispatcher is injected into the command handler

config/common.yml
L3RAZ: Yes, I guess so. But I would need to make some tests, because this code comes from as far as PS 1.7 and probably there wasn't a command handler back then

src/PayPal/Sdk/PayPalSdkConfigurationBuilder.php
```diff
@@ -47,70 +47,18 @@ class PayPalSdkConfigurationBuilder
...
-    }
+        private Env $env,
+        private PayPalConfiguration $configuration,
+        private PayPalPayLaterConfiguration $payLaterConfiguration,
+        private FundingSourceConfigurationRepository $fundingSourceConfigurationRepository,
+        private ExpressCheckoutConfiguration $expressCheckoutConfiguration,
+        private ShopContext $shopContext,
+        private PrestaShopContext $prestaShopContext,
+        private LoggerInterface $psCheckoutLogger,
+        private  FundingSourceEligibilityConstraint $fundingSourceEligibilityConstraint,
```
seiwan, line 59:
```suggestion
        private FundingSourceEligibilityConstraint $fundingSourceEligibilityConstraint,
```

seiwan (changes requested):

controllers/front/DispatchWebHook.php
```diff
@@ -283,7 +283,7 @@ protected function displayRestrictedCountryPage()
      *
      * @return false
      */
-    protected function geolocationManagement($default_country)
+    protected function geolocationManagement($defaultCountry)
```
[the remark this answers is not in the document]
L3RAZ: $defaultCountry is the correct name

.github/workflows/php.yml
```diff
@@ -33,21 +34,21 @@ jobs:
       - run: composer install
 
       - name: Run Header Stamp in Dry Run mode
-        run: php vendor/bin/header-stamp --license=vendor/prestashop/header-stamp/assets/afl.txt --exclude=.github,node_modules,vendor,tests,_dev --dry-run
+        run: php vendor/bin/header-stamp --license=vendor/prestashop/header-stamp/assets/afl.txt --exclude=.github,vendor,tests --dry-run
 
   php-linter:
-    name: PHP Syntax check 7.2|7.3|7.4
+    name: PHP Syntax check 8.1|8.2|8.2
```
PSebou, line 40: maybe we can just have 8.1|8.2 or 8.1|8.2|8.3

L3RAZ: Yes, I added a typo, it should be 8.1|8.2|8.3

seiwan (changes requested):

controllers/admin/AdminPaypalOnboardingPrestashopCheckoutController.php
```diff
@@ -122,7 +122,7 @@ public function init()
     /**
      * {@inheritdoc}
      */
-    protected function isAnonymousAllowed()
+    public function isAnonymousAllowed()
```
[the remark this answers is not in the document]
L3RAZ: Yes, in PS9 AdminControllerCore class it's public

ps_checkout.php
```diff
@@ -101,8 +101,8 @@ class Ps_checkout extends PaymentModule
      * @var \PrestaShop\ModuleLibServiceContainer\DependencyInjection\ServiceContainer
      */
     private $serviceContainer;
-    private static $merchantIsValid;
-    private static $currencyIsAllowed;
+    private ?bool $merchantIsValid;
```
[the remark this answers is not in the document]
L3RAZ: The whole point is to not check if merchant is valid multiple times. If it's null - we set the value. If it's bool - we know if it's valid or not

controllers/admin/AdminAjaxPrestashopCheckoutController.php
```diff
@@ -1067,4 +1063,12 @@ public function ajaxProcessDownloadLogs()
...
         exit;
     }
+
+    public function ajaxDie($value = null, $controller = null, $method = null)
+    {
+        header('Cache-Control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0');
+        header('Content-type: application/json');
+        echo json_encode($value);
+        exit;
+    }
```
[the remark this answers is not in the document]
L3RAZ: We would also need to modify BO SDK for it, because the expected response is true/false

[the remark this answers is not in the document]
Matt75: Are we handle the response content on BO-SDK? Or we rely only on http status code?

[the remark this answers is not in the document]
L3RAZ: We check response content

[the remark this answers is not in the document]
Matt75: Ok let's keep it like that...

Matt75 (approved):

Matt75 (approved):

matks (commented):
I approve commits
- https://github.com/PrestaShopCorp/ps_checkout/pull/1304/commits/fb006750d675e1d8f616fd1763c5792be586a5c3
- https://github.com/PrestaShopCorp/ps_checkout/pull/1304/commits/6e46e8f3ebcd57ff8359b819beeff56ad7c79352

But as I'm not a checkout expert I prefer to wait for a 2nd review ;)

Matt75 (approved):

PR #1389, opened 2025-10-04
Cast cart->id_customer to int before calling getFundingSourceTokens, …

zelarg:
…as in third party checkout context, id_customer may be empty on initial checkout form load.

## Self-Checks

## Summary

In third party checkout, when payment methods are called with no user, cart->id_customer will be empty, which would result in internal server error when calling typed-parameter method public function getFundingSourceTokens(int $customerId)

In previous ps_checkout module versions, cart->id_customer was casted to int in the FundingSourceProvider.php, with: 
```
public function getSavedTokens($customerId)
    {
        if ((int) $customerId && $this->payPalConfiguration->isVaultingEnabled()) {
```

## QA Checklist Labels
- [x] Regression (third party checkout module support)

<br><br>

It's very simple code change without any visual impact and with no impact to standard implementation in the native Prestashop checkout, it just adds more resilience and support for third party checkout modules.
- [x] Tested on dev store and also in production.

### DIFF
--- a/ps17/ps_checkout.php
+++ b/ps17/ps_checkout.php
@@ -680,7 +680,7 @@ class Ps_Checkout extends PaymentModule
 
         $paymentOptions = [];
 
-        foreach ($fundingSourceTokenPresenter->getFundingSourceTokens($cart->id_customer) as $fundingSource) {
+        foreach ($fundingSourceTokenPresenter->getFundingSourceTokens((int) $cart->id_customer) as $fundingSource) {
             $paymentOptions[] = $fundingSource->getName();
         }
 
--- a/ps8/ps_checkout.php
+++ b/ps8/ps_checkout.php
@@ -680,7 +680,7 @@ class Ps_Checkout extends PaymentModule
 
         $paymentOptions = [];
 
-        foreach ($fundingSourceTokenPresenter->getFundingSourceTokens($cart->id_customer) as $fundingSource) {
+        foreach ($fundingSourceTokenPresenter->getFundingSourceTokens((int) $cart->id_customer) as $fundingSource) {
             $paymentOptions[] = $fundingSource->getName();
         }
 
--- a/ps9/ps_checkout.php
+++ b/ps9/ps_checkout.php
@@ -679,7 +679,7 @@ class Ps_Checkout extends PaymentModule
 
         $paymentOptions = [];
 
-        foreach ($fundingSourceTokenPresenter->getFundingSourceTokens($cart->id_customer) as $fundingSource) {
+        foreach ($fundingSourceTokenPresenter->getFundingSourceTokens((int) $cart->id_customer) as $fundingSource) {
             $paymentOptions[] = $fundingSource->getName();
         }
 


matks (commented):
Hello @zelarg and thank you for the PR

> In third party checkout, when payment methods are called with no user, cart-> id_customer will be empty

I understand it's better to write code against this usecase but... isn't the usecase "payment method called with no user" useless?

If there is no user, why payment methods are shown 😉 ?

zelarg:
> If there is no user, why payment methods are shown 😉 ?

Because we want to show payment (and shipping) options to the visitor without requesting any data from them.
This is how (almost) all third party checkout modules work, so there's a demand on the market for such functionality.

I will reverse the question - why does payment module need customer->id to show its available payment option(s)? Apparently other ecommerce solutions (e.g. Shopify) do not ask for user account to show CC form to visitors.

matks:
> I will reverse the question - why does payment module need customer->id to show its available payment option(s)?

Because the payment options presented depend on customer situation

If customer is in Spain, or in Mexico, or in China, or in Thailand, the payment options available are different, just like shipping options

Also some payment options might depend on customer status: guest (1st order)? existing customer (2nd order)? specific group?

In PrestaShop the mechanism is "tell me who is your customer, I will tell you what are the available payment and shipping options I have for her/him"

matks:
> Apparently other ecommerce solutions (e.g. Shopify) do not ask for user account to show CC form to visitors.

If you present all possible payment options to customers before knowing them
then they identify themselves
you might disappoint them because you "remove" something they wanted

For example on your shop you say "We accept payments using PayPal, Apple Pay, Bank transfer and Card"
but when customer says he's from France you remove Apple Pay because the payment option was only available for US customers
it's a negative scenario :/

zelarg:
> but when customer says he's from France you remove Apple Pay because the payment option was only available for US customers
> it's a negative scenario :/

Correct, there are some border cases, mitigations are: 
- there's a default country set in Prestashop (and also shop owner can set payment methods visibility to 'Visitor' group)
- once user is presented with checkout form, the default country is pre-selected in his shipping address and shipping/payment options are shown for this country
- if user changes the country, this is updated in the address and payment options are re-fetched.

This is not just a plain theory, it works this way just fine for years and I cannot remember if I ever heard about such complaint from clients (such cases simply doesn't occur, the mechanism to mitigate this behavior already works). However, when certain payment method shall be visible, but is missing - that's quite often the source of complaints (typically when payment method has any pre-requirements to output anything in paymentOption hook)

> In PrestaShop the mechanism is "tell me who is your customer, I will tell you what are the available payment and shipping options I have for her/him"

This is perfectly valid and I understand it's set-up this way for a long time, but shouldn't payment methods display logic depend on PS core/hook execution level and not decided in particular paymentOption hook (can we agree on this?)

All I'm asking for is - do not change how ps_checkout works in the native checkout, just make a minor modification, so that life of third party checkout module developers and their clients gets easier. We do these modifications in ps_checkout files anyway (there's not workaround to them AFAIK), but with every update they get erased and have to be re-applied.

Maofree:
Hi
I've this problem on two websites, using the module The Checkout

Matt75 (dismissed):

matks:
Hello @zelarg we are preparing PS Checkout release 5.0.6 so we're going to merge your code in it 😉 

I backported your PR in https://github.com/PrestaShopCorp/ps_checkout/pull/1403 because you target the `main` branch while we need the `5.0.x` branch. I cherry-picked your commit so you remain the legit author of the bugfix.

I close this PR then -> replaced by https://github.com/PrestaShopCorp/ps_checkout/pull/1403

PR #1403, opened 2025-11-05
Cast cart->id_customer to int before calling getFundingSourceTokens

matks:
Cast cart->id_customer to int before calling getFundingSourceTokens as in third party checkout context, id_customer may be empty on initial checkout form load.

## Self-Checks

## Summary

In third party checkout, when payment methods are called with no user, cart->id_customer will be empty, which would result in internal server error when calling typed-parameter method public function getFundingSourceTokens(int $customerId)

In previous ps_checkout module versions, cart->id_customer was casted to int in the FundingSourceProvider.php, with: 
```
public function getSavedTokens($customerId)
    {
        if ((int) $customerId && $this->payPalConfiguration->isVaultingEnabled()) {
```

## QA Checklist Labels
- [x] Regression (third party checkout module support)

<br><br>

It's very simple code change without any visual impact and with no impact to standard implementation in the native Prestashop checkout, it just adds more resilience and support for third party checkout modules.
- [x] Tested on dev store and also in production.

### DIFF
--- a/ps17/ps_checkout.php
+++ b/ps17/ps_checkout.php
@@ -680,7 +680,7 @@ class Ps_Checkout extends PaymentModule
 
         $paymentOptions = [];
 
-        foreach ($fundingSourceTokenPresenter->getFundingSourceTokens($cart->id_customer) as $fundingSource) {
+        foreach ($fundingSourceTokenPresenter->getFundingSourceTokens((int) $cart->id_customer) as $fundingSource) {
             $paymentOptions[] = $fundingSource->getName();
         }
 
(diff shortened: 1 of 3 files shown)


matks:
Backport of https://github.com/PrestaShopCorp/ps_checkout/pull/1389 for 5.0.x

matks:
Thank you @zelarg

matks:
Hello we finally released you bugfix @zelarg thanks!

It was late because we wanted to fix a nasty and complex bug in 5.0.6

PR #1436, opened 2025-12-18
[PAYSHIP-3719] Reauthorize order authorization

gaelreyrol:
## Self-Checks
- [ ] I have performed a self-review of my code.
- [ ] I have updated/added necessary technical documentation in the [README](/README.md) file.

## JIRA task link

Resolves: [PAYSHIP-3719](https://forge.prestashop.com/browse/PAYSHIP-3719)

## QA Checklist Labels
- [ ] Bug fix? 
- [x] New feature? 
- [ ] Improvement? 
- [ ] Technical debt? 
- [ ] Covered by tests? 

## Frontend Changes

<br><br>

[!] Don't forget to include JIRA task number in PR name or branch name. [!]

[!] Don't forget to update *changelog.md* with new changes. [!]

- [ ] Verify if the version is correct.
- [ ] Update demo environments used for testing.
- [ ] Ensure changelog contains all changes.
- [ ] Update documentation if needed.
- [ ] Test if everything works properly.


.github/workflows/lint.yml
```diff
@@ -1,15 +1,27 @@
...
-  prestashop:
-    name: ${{ matrix.ps-version }}
+  php-cs-fixer:
+    name: PHP CS Fixer
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.5'
```
matks, line 18: I'm wondering if using PHP8.5 for PHP CS Fixer means we could apply some codestyle not compatible with PHP8.1 or PHP7 and break something for one of our target builds (PS 1.7, PS 8, PS 9)

I just ask, maybe I am paranoid

.php-cs-fixer.php
```diff
@@ -1,15 +1,27 @@
 <?php
 
-$config = new PhpCsFixer\Config();
+declare(strict_types=1);
```
matks, line 3: 👍

api/src/Dto/PayPal/LinkDescription.php
```diff
@@ -0,0 +1,107 @@
+<?php
+
+namespace PsCheckout\Api\Dto\PayPal;
+
+/**
+ * The request-related [HATEOAS link](/api/rest/responses/#hateoas-links) information.
```
matks: You need to add the AFL license header, just like every other code files

```
 * Copyright since 2007 PrestaShop SA and Contributors
 * PrestaShop is an International Registered Trademark & Property of PrestaShop SA
 ...
 ...
```

api/src/Dto/PayPal/LinkDescription.php
```diff
@@ -0,0 +1,107 @@
...
+    private $rel;
+
+    /**
+     * @var string|null
+     */
+    private $method;
+
+    /**
+     * @param string $href
+     * @param string $rel
```
matks, line 42: Missing 3rd parameter in phpdoc

api/src/Dto/PayPal/LinkDescription.php
```diff
@@ -0,0 +1,107 @@
...
+     * org/assignments/link-relations/link-relations.xhtml).
+     */
+    public function getRel(): string
+    {
+        return $this->rel;
+    }
+
+    /**
+     * Sets Rel.
+     * The [link relation type](https://tools.ietf.org/html/rfc5988#section-4), which serves as an ID for a
```
matks, line 92: Mmmm ... this might not be very very useful 😄 ?

api/src/Dto/PayPal/Money.php
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * @var string
+     */
+    private $value;
+
+    /**
+     * @param string $currencyCode
+     * @param string $value
+     */
+    public function __construct(string $currencyCode, string $value)
```
matks, line 39: - I guess it's inspired by https://www.moneyphp.org/ ?
- I wonder if we could add some validation on the currencyCode (enuuuuuuum 😄 is what I think about but I don't know if it applies well here)

api/src/Dto/PayPal/Payment/AuthorizationLinkRelation.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace PsCheckout\Api\Dto\PayPal\Payment;
+
+/**
+ * The status for the authorized payment.
+ */
+class AuthorizationLinkRelation
```
matks, line 26: Aaaaah ❤️ I love this

core/tests/Unit/PayPal/Order/Processor/UpdateExternalPayPalOrderProcessorTest.php
```diff
@@ -227,6 +229,8 @@ public function testItUpdatesOrderWhenChangesDetected(): void
 
     public function testItThrowsExceptionWhenUpdateFails(): void
     {
+        $this->markTestSkipped('This test is broken.');
```
matks, line 232: 😆

core/src/PayPal/Payment/Authorization/Action/ReauthorizeAuthorizationAction.php
```diff
@@ -0,0 +1,162 @@
...
+            $this->payPalOrderAuthorizationRepository->save($payPalReauthorization);
+        } catch (\Throwable $exception) {
+            $this->logger->error('PayPal Order reauthorization save failed', [
+                'exception' => $exception,
+                'order_id' => $payPalOrder->getId(),
+                'authorization_id' => $parentAuthorization->getId(),
+                'reauthorization_id' => $reauthorization->getId(),
+            ]);
+
+            throw new PsCheckoutException(
```
matks, line 161: I don't know if we should throw the generic PsCheckoutException here or create more subclasses. Looking for more opinions

matks (commented):
This is globally good in my opinion

One warning to be aware of :

On the core of PrestaShop https://github.com/PrestaShop/prestashop we also had a lot of big arrays in the past. We started to use DTOs instead, slowly. We started getting bug reports like https://github.com/PrestaShop/PrestaShop/issues/22215

What happened is that, since arrays are very flexible, some people had wrong / corrupted data in their database
but the code was somehow "working" (= the expected behavior was working)
and when they upgraded their software with the latest version that replaced arrays with DTOs, suddenly "their store was broken"

What was happening is that the page / behavior previously "working" were now failing because our code was more strict because of the DTOs

The conclusion is: using DTOs is good and we should start doing it but we have to anticipate that the use of arrays for multiple years have created a situation where some stores / modules are working despite the data being wrong or corrupted, and the upgrade to more strict code will make them fail. From their point of view it will look like "it worked before, it is broken now"

.github/workflows/lint.yml
gaelreyrol: It should not, PHP-CS-Fixer acts from a set of rules, if those rules do not contain PHP 8.5 code, it won't apply incompatible code style.

api/src/Dto/PayPal/LinkDescription.php
gaelreyrol: It's extracted from PayPal SDK, I did not wanted to spent time to remove those kind of comments.

api/src/Dto/PayPal/Money.php
gaelreyrol: No it's not, it's extracted from PayPal. Also those are DTOs, not Domain classes, therefore they should not contain any domain logic such as currency code validation. They are here to map the data from PayPal into POPO only.

core/tests/Unit/PayPal/Order/Processor/UpdateExternalPayPalOrderProcessorTest.php
gaelreyrol: Otherwise, running unit tests fails, which makes it tricking to test your own code when polluted by errors from other tests. I made a PR to fix those tests a long ago: https://github.com/PrestaShopCorp/ps_checkout/pull/1426

core/src/PayPal/Payment/Authorization/Action/ReauthorizeAuthorizationAction.php
gaelreyrol: It's what I think, we should use exception inheritance instead of using a constant code to have exceptions variations but on step at a time. I did not wanted to change the logic on this part yet.

gaelreyrol:
> This is globally good in my opinion
> 
> One warning to be aware of :
> 
> On the core of PrestaShop https://github.com/PrestaShop/prestashop we also had a lot of big arrays in the past. We started to use DTOs instead, slowly. We started getting bug reports like [PrestaShop/PrestaShop#22215](https://github.com/PrestaShop/PrestaShop/issues/22215)
> 
> What happened is that, since arrays are very flexible, some people had wrong / corrupted data in their database but the code was somehow "working" (= the expected behavior was working) and when they upgraded their software with the latest version that replaced arrays with DTOs, suddenly "their store was broken"
> 
> What was happening is that the page / behavior previously "working" were now failing because our code was more strict because of the DTOs
> 
> The conclusion is: using DTOs is good and we should start doing it but we have to anticipate that the use of arrays for multiple years have created a situation where some stores / modules are working despite the data being wrong or corrupted, and the upgrade to more strict code will make them fail. From their point of view it will look like "it worked before, it is broken now"

What matters is the data, if your data change, your DTOs or your arrays will break. The responsibility of DTO is not to guarantee that you will always be able to map raw data to POPOs, it is to have an anti-corruption layer between an external system and your domain. You don't what to play with unmapped / arbitrary data in your domain, you want to be able to map and validate them when the data is received so you can apply domain logic safely. So when mapping/validating external data, if it breaks, it allows you to recover from it and not have your domain to react badly to it, it's the point of it, precisely.

I understand the warning here, but having it to break when mapping the data is preferable and more predictable. It also allows you to have cleaner and more robust domain code.

Also what you pointed out is about a query, a query is not a DTO, it's part of your domain which should indeed always be valid.

matks:
@gaelreyrol When I mention DTO, I mean the idea of using objects instead of array. In don't distinguish DTOs from POPOs here but it seems you do so it probably leads to misunderstandings.

> I understand the warning here, but having it to break when mapping the data is preferable and more predictable

More predictable, yes, preferable, no. Customers prefer unreliable and legacy code that is providing them value rather than clean and modern code that provides them no value. We cannot simply say "the new code is better, so I understand it breaks your shop and the old code did not break it but the new code is better", they don't care.

That is why I say: `we have to anticipate`. The usecase I mentioned will happen. There will be a merchant that is using today PS Checkout with corrupted data and "it works" because arrays are permissive, he will upgrade to the latest version of PS Checkout and it will cause a fatal error and prevent him from receiving payments. What can we do to mitigate this? Can we display an error page with a clear explanation of what happened and how to get help? Can we inform the merchant of what is the corrupted data, why PS Checkout cannot work with it, how to fix it? Can we offer alternatives?

It does not have to be something complex. For example (it's just an idea):
- wrap every piece of code where we hydrate a POPO from the database in a try/catch
- catch any mapping error and throw a specific exception, for example PsCheckoutCannotMapToPOPOException
- also create an error log of this with a specific tag so we can easily identify it when investigating Support tickets
- catch this exception at controller level, and present a pop-in that is presented with the "I got a PsCheckoutCannotMapToPOPOException Help me!" button that leads to a page we create in PS Checkout wiki https://github.com/PrestaShopCorp/ps_checkout/wiki

Now if I am not wrong, what I am saying here is irrelevant to this PR, because this is a new feature so there will be no pre-existing data. I wanted to talk about global usage of POPO/DTO instead of arrays 😉 .

gaelreyrol:
@matks To be fair, I think you are wrong and that what you say is irrelevant to this PR. I know from a long time what it means to use POPO/DTO and how to handle it safely, unlike arrays which are a door to hell.

> More predictable, yes, preferable, no. Customers prefer unreliable and legacy code that is providing them value rather than clean and modern code that provides them no value. We cannot simply say "the new code is better, so I understand it breaks your shop and the old code did not break it but the new code is better", they don't care.

This is absolutely not what I said, the customer does not care about better code indeed but WE do and when investigating it matters a lot. So at least when it breaks on the customer side, having clear exceptions instead of illegal offsets is a benefit.

> That is why I say: we have to anticipate. The usecase I mentioned will happen. There will be a merchant that is using today PS Checkout with corrupted data and "it works" because arrays are permissive, he will upgrade to the latest version of PS Checkout and it will cause a fatal error and prevent him from receiving payments. What can we do to mitigate this? Can we display an error page with a clear explanation of what happened and how to get help? Can we inform the merchant of what is the corrupted data, why PS Checkout cannot work with it, how to fix it? Can we offer alternatives?

You are putting the cart before the horse, we cannot achieve a clean recovery of errors with the current implementation, for now the only way is to propagate this damn too large `PsCheckoutException`. By the way, if you look closely, my implementation is full of try/catch to prevent any uncatched exceptions.

PS: 

> There will be a merchant that is using today PS Checkout with corrupted data and "it works" because arrays are permissive

Where the heck an array is permissive, it is not, the PHP type system is, which allows to handle value types at runtime. This is absolutely different.

ps17/composer.json
```diff
@@ -13,16 +10,18 @@
...
         "gmponos/guzzle_logger": "^2.2",
+        "phpdocumentor/reflection-docblock": "^5.3",
+        "prestashop/module-lib-guzzle-adapter": "^1.0",
         "prestashop/module-lib-mbo-installer": "^2.0",
         "prestashop/module-lib-service-container": "^2.0",
         "prestashop/prestashop-accounts-installer": "^1.0",
-        "prestashop/module-lib-guzzle-adapter": "^1.0",
-        "giggsey/libphonenumber-for-php": "^8.12"
+        "symfony/property-access": "^2.8",
+        "symfony/serializer": "^2.8"
```
Matt75: Already bundled in PrestaShop Core dependencies so it will be available inside PrestaShop for the module. If you need it for tests or local, you can move it to dev dependencies instead.

ps8/composer.json
```diff
@@ -1,28 +1,24 @@
...
         "gmponos/guzzle_logger": "^2.2",
+        "phpdocumentor/reflection-docblock": "^5.3",
+        "prestashop/module-lib-guzzle-adapter": "^1.0",
         "prestashop/module-lib-mbo-installer": "^2.0",
         "prestashop/module-lib-service-container": "^2.0",
         "prestashop/prestashop-accounts-installer": "^1.0",
-        "prestashop/module-lib-guzzle-adapter": "^1.0",
-        "giggsey/libphonenumber-for-php": "^8.12"
+        "symfony/property-access": "^4.4",
+        "symfony/serializer": "^4.4"
```
Matt75: Already bundled in PrestaShop Core dependencies so it will be available inside PrestaShop for the module. If you need it for tests or local, you can move it to dev dependencies instead.

ps9/composer.json
```diff
@@ -1,28 +1,24 @@
...
         "gmponos/guzzle_logger": "^2.2",
+        "phpdocumentor/reflection-docblock": "^5.6",
+        "prestashop/module-lib-guzzle-adapter": "^1.0",
         "prestashop/module-lib-mbo-installer": "^2.0",
         "prestashop/module-lib-service-container": "^2.0",
         "prestashop/prestashop-accounts-installer": "^1.0",
-        "prestashop/module-lib-guzzle-adapter": "^1.0",
-        "giggsey/libphonenumber-for-php": "^8.12"
+        "symfony/property-access": "^6.4",
+        "symfony/serializer": "^6.4"
```
Matt75: Already bundled in PrestaShop Core dependencies so it will be available inside PrestaShop for the module. If you need it for tests or local, you can move it to dev dependencies instead.

Matt75 (approved):

core/src/PayPal/Payment/Authorization/Action/ReauthorizeAuthorizationAction.php
```diff
@@ -0,0 +1,180 @@
...
+                sprintf('PayPal Order %s re-authorization database failure for authorization %s', $payPalOrder->getId(), $reauthorization->getId()),
+                PsCheckoutException::PAYPAL_AUTHORIZATION_DATABASE_FAILURE,
+                $exception
+            );
+        }
+
+        if (in_array($reauthorization->getStatus(), [PayPalAuthorizationStatus::CREATED, PayPalAuthorizationStatus::PENDING], true)) {
+            $this->logger->info(sprintf('PayPal Order %s re-authorization %s is %s', $payPalOrder->getId(), $reauthorization->getId(), $reauthorization->getStatus()));
+            $this->paymentPendingEventHandler->handle($payPalOrder);
```
L3RAZ, line 170: @gaelreyrol Just viewed what's done in this PR and we absolutely don't want to call paymentPendingEventHandler, because it attempts to create a PrestaShop order with pending payment order state.

L3RAZ: Maybe call SetPendingOrderStateAction alone, but not the event handler

ps9/config/front/process.yml
```diff
@@ -269,4 +269,13 @@ services:
       - '@PsCheckout\Core\PayPal\Order\Handler\PaymentPendingEventHandler'
       - '@PsCheckout\Core\PayPal\Order\Handler\PaymentDeniedEventHandler'
       - '@PsCheckout\Core\PayPal\Order\Provider\PayPalOrderProvider'
-      - '@PsCheckout\Infrastructure\Repository\PayPalOrderAuthorizationRepository'
\ No newline at end of file
+      - '@PsCheckout\Infrastructure\Repository\PayPalOrderAuthorizationRepository'
+
+  PsCheckout\Core\PayPal\Payment\Authorization\Action\ReauthorizeAuthorizationAction:
```
L3RAZ, line 274: Should be in shared process config file, because it needs to be accesible by back office container

PR #1469, opened 2026-02-26
Fix fatal error when purchase unit has no items key

louisbels:
## Description

`CreatePayPalOrderProcessor::savePurchaseUnits()` and `UpdatePayPalOrderPurchaseUnitAction::execute()` crash with a fatal error when PayPal returns a purchase unit without an `items` key.

### Error

```
PHP Warning: Undefined array key "items" in CreatePayPalOrderProcessor.php on line 223

PHP Fatal error: Uncaught TypeError: PayPalOrderPurchaseUnit::__construct():
Argument #4 ($items) must be of type array, null given
```

## Root cause

`$purchaseUnit['items']` is accessed without checking the key exists. When PayPal omits `items` from the response, `null` is passed to the constructor which expects `array`.

## Fix

Use null coalescing operator to default to an empty array:

```php
$purchaseUnit['items'] ?? []
```

Applied in both:
- `core/src/PayPal/Order/Processor/CreatePayPalOrderProcessor.php`
- `core/src/PayPal/Order/Action/UpdatePayPalOrderPurchaseUnitAction.php`

## Environment

- ps_checkout version: 8.5.1.0 (v5.1.0)
- PrestaShop: 8.x
- PHP: 8.1

### DIFF
--- a/core/src/PayPal/Order/Action/UpdatePayPalOrderPurchaseUnitAction.php
+++ b/core/src/PayPal/Order/Action/UpdatePayPalOrderPurchaseUnitAction.php
@@ -74,7 +74,7 @@ class UpdatePayPalOrderPurchaseUnitAction implements UpdatePayPalOrderPurchaseUn
                 $orderResponse->getId(),
                 crc32(json_encode($purchaseUnit)),
                 $purchaseUnit['reference_id'],
-                $purchaseUnit['items']
+                $purchaseUnit['items'] ?? []
             );
 
             $this->payPalOrderPurchaseUnitRepository->save($payPalPurchaseUnit);
--- a/core/src/PayPal/Order/Processor/CreatePayPalOrderProcessor.php
+++ b/core/src/PayPal/Order/Processor/CreatePayPalOrderProcessor.php
@@ -220,7 +220,7 @@ class CreatePayPalOrderProcessor implements CreatePayPalOrderProcessorInterface
                 $orderResponse->getId(),
                 crc32(json_encode($purchaseUnit)),
                 $purchaseUnit['reference_id'],
-                $purchaseUnit['items']
+                $purchaseUnit['items'] ?? []
             );
 
             $this->payPalOrderPurchaseUnitRepository->save($payPalPurchaseUnit);


matks (commented):
Hello @louisbels interesting PR 👍 thanks for submitting it

A paypal response without `items` seems weird but I guess if you submitted this PR it is because you have experienced some usecases where it happens. Do you have some infos about which cases?

We would love to be able to reproduce the bug

louisbels:
Hi @matks, thanks for the review!

Yes, this is happening in production on a PrestaShop 8.x / PHP 8.1 store with ps_checkout 8.5.1.0. We've logged **58 occurrences over 5 days**, affecting multiple different customers. Here's what we see in nginx error logs:

```
PHP Warning: Undefined array key "items" in
  .../CreatePayPalOrderProcessor.php on line 223

PHP Fatal error: Uncaught TypeError: PayPalOrderPurchaseUnit::__construct():
  Argument #4 ($items) must be of type array, null given
```

Cross-referencing with the access logs gives us more context:

- All failing requests return **HTTP 500 with 0 bytes** response body on `POST /module/ps_checkout/create`
- The bug is **intermittent**: the same customer with the same token gets multiple 500s, then eventually succeeds (200) on a subsequent retry. For example, one customer got three consecutive 500s before a successful 200 about 30 seconds later.
- It's not browser-specific (Firefox, Chrome, Safari all affected)
- Multiple different customers can hit it at the same time

This suggests PayPal intermittently omits the `items` key from the purchase unit in its response. Unfortunately the nginx logs only capture PHP stderr, not the actual PayPal API response body, so we can't see exactly what PayPal returned.

The fix is intentionally minimal — just defaulting to an empty array with `?? []` — so it handles the missing key gracefully without changing any other behavior.

matks:
> The fix is intentionally minimal — just defaulting to an empty array with ?? [] — so it handles the missing key gracefully without changing any other behavior.

Yes we will probably merge this in next patch version because it cannot hurt 😉 

But on our side we need to find out when and how the `items` can be omitted
- to add automated testing to reproduce the problem and make sure it does not happen in the future
- and because it could the tree hiding the forest: I want to understand what's going on

matks (approved):
Hello again @louisbels we analyzed this PR together with @Matt75 and think it'll help 😉 

Regarding the root cause, @Matt75 believes it could happens with Express Checkout on product page. We noticed a bug yesterday when sometimes we cannot retrieve the quantity_wanted data from the theme, so the "add to cart" action fails and in this scenario it could happen we try to create a PayPal Order from an empty cart

The quantity_wanted missing data should be an issue on the Core so far, but it's random so hard to be sure

Anyway the code changes you submitted are probably a positive addition 👍 

I see you targeted branch`main` which is not suited
but I think I will simply cherry-pick your changes into next patch version

Matt75 (approved):

matks:
Hello @louisbels your fix was included in the v5.2.0 release published yesterday https://github.com/PrestaShopCorp/ps_checkout/releases/tag/v5.2.0

PR #1499, opened 2026-04-02
Removed sentry and it's references

L3RAZ:


gaelreyrol (commented):
The composer.lock(s) don't need to be updated?

L3RAZ:
@gaelreyrol Seems like they never were updated, only composer.json was

gaelreyrol:
@L3RAZ Okey, we will see if the zip creation don't fail!

gaelreyrol (dismissed):

gaelreyrol (approved):

PR #1509, opened 2026-04-03
[PAYSHIP-3996] Fix order stuck in Partial payment after full multi-capture

gaelreyrol:
## Summary

- Fix order status not transitioning from "Partial payment" to "Payment accepted" when multiple partial captures sum to the full authorization amount
- Replace `hasBeenPaid()` guard with a check against the COMPLETED state only, allowing partially-paid orders to be re-evaluated on subsequent capture webhooks
- Sum all captures using `getCaptures()` + `array_reduce()` instead of comparing only the first capture via `getCapture()`

## Related

- [PAYSHIP-3996](https://forge.prestashop.com/browse/PAYSHIP-3996)

## Test plan

- [x] Create an order with AUTHORIZE intent
- [x] Perform a first partial capture (amount < order total) → order should transition to "Partial payment"
- [x] Perform a second capture for the remaining amount → order should transition to "Payment accepted"
- [x] Verify a single full capture still transitions directly to "Payment accepted"
- [x] Verify an already-completed order is not re-processed on duplicate webhooks
- [x] Run `make phpstan` for all MODULE_VERSION values
- [x] Run `make integration-test`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/src/OrderState/Action/SetCompletedOrderStateAction.php
+++ b/core/src/OrderState/Action/SetCompletedOrderStateAction.php
@@ -25,6 +25,7 @@ use PsCheckout\Core\Order\Validator\OrderAmountValidator;
 use PsCheckout\Core\Order\Validator\OrderAmountValidatorInterface;
 use PsCheckout\Core\OrderState\Configuration\OrderStateConfiguration;
 use PsCheckout\Core\OrderState\Service\OrderStateMapperInterface;
+use PsCheckout\Core\PayPal\Order\Configuration\PayPalCaptureStatus;
 use PsCheckout\Core\PayPal\Order\Provider\PayPalOrderProviderInterface;
 use PsCheckout\Core\PayPal\Order\Repository\PayPalOrderRepositoryInterface;
 use PsCheckout\Infrastructure\Repository\OrderRepositoryInterface;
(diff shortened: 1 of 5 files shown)


gaelreyrol:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Reviewed: guard change from `hasBeenPaid()` to state-ID comparison, capture amount summing via `array_reduce`, PHPStan baseline updates across ps8/ps9/ps17, and new multi-capture integration test.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

core/src/OrderState/Action/SetCompletedOrderStateAction.php
```diff
@@ -93,13 +94,28 @@ public function execute(string $payPalOrderId)
...
+        $completedStateId = $this->orderStateMapper->getIdByKey(OrderStateConfiguration::PS_CHECKOUT_STATE_COMPLETED);
+        $hasBeenCompleted = count($order->getHistory($order->id_lang, $completedStateId)) > 0;
+
+        if ($hasBeenCompleted) {
+            return;
+        }
+
+        $completedCaptures = array_filter($payPalOrderResponse->getCaptures(), function ($capture) {
+            return isset($capture['status']) && $capture['status'] === PayPalCaptureStatus::COMPLETED;
+        });
```
Matt75: You may want to broaden the capture-status filter when computing the “total captured” amount.

Possible capture statuses are `COMPLETED`, `DECLINED`, `PARTIALLY_REFUNDED`, `PENDING`, `REFUNDED`, `FAILED`, and a capture can transition from `COMPLETED` to `PARTIALLY_REFUNDED` or `REFUNDED`.

If we only sum captures currently in `COMPLETED`, we’ll undercount in scenarios where a capture was successfully completed and later refunded (partially or fully), because it would no longer be in `COMPLETED` even though it *was* captured.

Maybe we should instead:
- compute **gross captured** by including captures in `COMPLETED` / `PARTIALLY_REFUNDED` / `REFUNDED` (i.e., “was captured at some point”), and
- compute **net captured** separately by subtracting refunds (if that’s what’s needed for order state).

As-is, restricting to `COMPLETED` feels like it could produce incorrect totals and keep the order in `Partial payment` even after full multi-capture + subsequent refund transitions.

Matt75 (approved):

